PR #3, opened 2024-07-16
Update investigation model

upasanapattnaik-eq:
Update to Investigation model to include NBS_Act_Entity fields for F_PAGE_CASE update.

### DIFF
--- a/investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/repository/model/reporting/InvestigationReporting.java
+++ b/investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/repository/model/reporting/InvestigationReporting.java
@@ -87,6 +87,13 @@ public class InvestigationReporting {
     private String notificationAddTime;
     private String notificationRecordStatusCd;
     private String notificationLastChgTime;
+    private Long caseManagementUid;
+    private Long nacPageCaseUid;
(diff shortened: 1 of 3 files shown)


investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/repository/model/reporting/InvestigationReporting.java
```diff
@@ -87,6 +87,13 @@ public class InvestigationReporting {
     private String notificationAddTime;
     private String notificationRecordStatusCd;
     private String notificationLastChgTime;
+    private Long caseManagementUid;
+    private Long nacPageCaseUid;
```
alog-enquizit, line 91: Hope we do have a corresponding update to the liquibase scripts

alog-enquizit (approved):
lgtm

investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/repository/model/reporting/InvestigationReporting.java
upasanapattnaik-eq: I believe they were updated in the 6.0.15.1 PR for Liquibase but the model was not updated. This makes sure the columns are getting updated.

PR #6, opened 2024-07-24
Cndit 1584 : Adding null checks to the Data Mart processing logic

alog-enquizit:
Adding null checks to the Data Mart processing logic
Updated and ran all the unit tests


upasanapattnaik-eq (approved):
LGTM! Thank you for the speedy update!

PR #7, opened 2024-07-24
CNDIT-1589: Liquibase migration to the NEDSS-DataReporting repository

alog-enquizit:
CNDIT-1589: Liquibase migration to the NEDSS-DataReporting repository


alog-enquizit:
Liquibase service image successfully built and image version successfully updated in the DTS1

liquibase-service/src/main/resources/application.yaml
```diff
@@ -0,0 +1,17 @@
+spring:
+  application:
+    name: liquibase-service
+  datasource:
+    password: ${DB_PASSWORD:-fake.fake.fake.1234}
+    username: ${DB_USERNAME:-sa}
+    url: ${DB_URL:localhost}
+    odse: ${DB_ODSE:NBS_ODSE}
+    rdb: ${DB_RDB:RDB}
```
upasanapattnaik-eq, line 9: Small note: Does this need to point to RDB_Modern here for internal testing or are we good to keep this as RDB_Modern?

liquibase-service/src/main/resources/db/odse/010-sp_investigation_event-001.sql
```diff
@@ -0,0 +1,2650 @@
+CREATE OR ALTER PROCEDURE dbo.sp_public_health_case_fact_datamart_event @phc_id_list nvarchar(max)
```
upasanapattnaik-eq, line 1: The investigation stored proc is not called here but it can be taken up in a different PR

upasanapattnaik-eq (approved):
LGTM!

liquibase-service/src/main/resources/application.yaml
alog-enquizit: Yeah we can control it locally by overriding this parameter. In prod deployment, we can override this in the ArgoCd

PR #8, opened 2024-07-26
Liquibase: sproc updates with bug fixes

upasanapattnaik-eq:
Updates to stored procedures:
+ Including bug fix versions for ODSE pre-processing sprocs- trim/substring handling for address fields.
+ Updates hepatitis stored procedure to pull page builder inv dimensions columns that were being missed.
+ Addition of `with (nolock)` to table joins for multiple/heavy transaction handling. 
+ Correct sproc for investigation pre-processing.

### DIFF
--- a/liquibase-service/src/main/resources/db/rdb/032-create_nrt_datamart_metadata-001.sql
+++ b/liquibase-service/src/main/resources/db/rdb/032-create_nrt_datamart_metadata-001.sql
@@ -1,14 +1,27 @@
-USE RDB_MODERN;
 IF NOT EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_datamart_metadata' and xtype = 'U')
-CREATE TABLE dbo.nrt_datamart_metadata (
-                                           condition_cd varchar(20) NOT NULL,
-                                           condition_desc_txt varchar(300) NULL,
-                                           Datamart varchar(18) NOT NULL,
-                                           Stored_Procedure varchar(36) NOT NULL
-                                       );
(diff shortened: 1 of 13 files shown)


liquibase-service/src/main/resources/db/rdb/032-create_nrt_datamart_metadata-001.sql
```diff
@@ -1,14 +1,27 @@
...
+    );
+    INSERT INTO dbo.nrt_datamart_metadata
+        (condition_cd,condition_desc_txt,Datamart,Stored_Procedure)
+    SELECT condition_cd,
+           condition_desc_txt,
+           'Hepatitis_Datamart',
+           'sp_hepatitis_datamart_postprocessing'
+    FROM NBS_SRTE.[dbo].[Condition_code]
+    WITH (NOLOCK)
+    WHERE CONDITION_CD IN ( '10110'
```
alog-enquizit, line 17: We are good with this condition codes right.

alog-enquizit (approved):
LGTM! Thanks for getting this!
👍

liquibase-service/src/main/resources/db/rdb/032-create_nrt_datamart_metadata-001.sql
upasanapattnaik-eq: Yep, these are specific in the hepatitis datamart. As a next step, we can identify the codes/requirements associated to the next datamarts being built and update this logic.

PR #12, opened 2024-08-02
CNDIT-1554: Send message to DLQ for empty or not existed database objects

sveselev:
## Description
Implement and throw a custom exception when the requested database object is unavailable on the database side (sp returns empty data for the specified uid). Send incoming message to DLT if empty data is returned from database.

- **NoDataException.java**: custom exception class;
- **Investigation**, **Observation**, **Organization**, **Person** services now throw the NoDataException in the `processMessage` listeners;
- Same logic is to be implemented in the context of upcoming **PostProcessing** and **LDF** services enhancement.

## JIRA
[CNDIT-1554](https://cdc-nbs.atlassian.net/browse/CNDIT-1554)


rshanmugam-enquizit (approved):
lgtm! ship it

PR #14, opened 2024-08-07
CNDIT-1492: DLQ implementation for Post Processing and LDF service

sveselev:
## Description
Add DLQ implementation for RTR post-processing and LDF pre-processing service. Used Investigation pre-processing DLQ implementation as the reference.

- Removed KafkaStreams usage from LDF pre-provcessing
- Added RetryableTopic and DLT strategy to LDF preprocessing and post-processing Kafka consumers
- Removed Investigation preprocessing dependency on PHC Fact Datamart sp execution success/error
- Adjusting unit tests, general code cleanups

## JIRA
[CNDIT-1492](https://cdc-nbs.atlassian.net/browse/CNDIT-1492)


investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
```diff
@@ -90,10 +90,14 @@ public String processInvestigation(String value) {
...
-                logger.info("Executing stored proc with ids: {} to populate PHС fact datamart", publicHealthCaseUid);
-                investigationRepository.populatePhcFact(publicHealthCaseUid);
-                logger.info("Stored proc executed");
+                try {
+                    // Calling sp_public_health_case_fact_datamart_event
+                    logger.info("Executing stored proc with ids: {} to populate PHС fact datamart", publicHealthCaseUid);
+                    investigationRepository.populatePhcFact(publicHealthCaseUid);
+                    logger.info("Stored proc executed");
+                } catch (Exception dbe) {
+                    logger.warn("Error processing PHC fact datamart: {}", dbe.getMessage());
```
upasanapattnaik-eq, line 99: 💯

upasanapattnaik-eq: Would this also go into the DLQ?

person-service/src/main/resources/application.yaml
```diff
@@ -17,7 +17,6 @@ spring:
       patientReporting:
         topic-name: ${PATIENT_REPORTING:nrt_patient}
     dlq:
```
upasanapattnaik-eq, line 19: For clarity- what would the format for the name of the dlq topic be?

sveselev: It's defined right after that:
```
    dlq:
      retry-suffix: _retry
      dlq-suffix: _dlt
```
So for the patient it's `nrt_person_dlt`

investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
sveselev: No, there is no topic for PHC, so we can't sent it to DLQ in that case. Otherwise entire investigation message would go to DLQ.

person-service/src/main/resources/application.yaml
upasanapattnaik-eq: Does the sink connector try to sink this into rdb_modern? The regex might try to pick it up, and it usually ends in an error if the table is not there.

sveselev: No, at least it's not supposed to be synchronized with db. DLQ is just for Kafka side of the process.

investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
upasanapattnaik-eq: In the case of an PHC datamart event erroring out, it might be a bit hard to find errors in publichealthcasefactdatamart. It is not the scope of this ticket, but I wonder if it would be more helpful to migrate it into the post-processing where the error can be put into a queue- or wherever we can keep the errors queued for this.

person-service/src/main/resources/application.yaml
upasanapattnaik-eq: The sink is currently listening for anything topic with the prefix nrt_{} in kafka. Not sure if it will discern this.

sveselev: I see the `"topics.regex": "nrt_[a-zA-Z_]+"`, and that regex will match dlt topics, indeed. If it was `nrt_[a-zA-Z]+` (no underscore included) then only `nrt_patient` would be matched, but not the `nrt_patient_dlt`.
Anyway, that's not a scope for this ticket, because the _dlt suffix was the initial solution for DLQ implemetation. This ticket just extended it for two remaining services, not covered initially.

upasanapattnaik-eq: Ah I see. That might be an issue when it gets autodeployed in DTS1. The regex currently accounts for tables like nrt_investigation_confirmation. 
There are a few options: 
- We could hardcode the sink connector's topic list with our post-processing expected set, but it does mean that the list needs to be maintained for all nrt tables that get sinked. 
- Could the dlq's prefix be updated?

upasanapattnaik-eq (approved):
LGTM! Thank you for the updates

ldfdata-service/src/main/resources/application.yaml
```diff
@@ -1,16 +1,18 @@
...
-          topic-name: nbs_state_defined_field_data
-      output:
-        ldfdata:
-          topic-name-reporting: nrt_ldf_data
+    input:
+      topic-name: nbs_state_defined_field_data
+    output:
+      topic-name-reporting: nrt_ldf_data
+    dlq:
+      retry-suffix: _retry
```
alog-enquizit, line 8: should this be `-retry` and `-dlt` as discussed

investigation-service/src/main/resources/application.yaml
```diff
@@ -9,7 +9,6 @@ spring:
       topic-name-observation: nrt_investigation_observation
       topic-name-notifications: nrt_notifications
     dlq:
-      topic-name-dlq: nrt_investigation_dlq
       retry-suffix: _retry
       dlq-suffix: _dlt
```
alog-enquizit, line 13: do we want to change here too from underscore to dash (`-retry` and `-dlt`)

ldfdata-service/src/main/resources/application.yaml
sveselev: DLQ and retry for `nrt_` topics is happening on the post-processing service, not in preprocessing, so we're good here

investigation-service/src/main/resources/application.yaml
sveselev: those suffixes applied to `nbs_` topics only in the pre-processing

alog-enquizit (approved):
LGTM 👍

PR #18, opened 2024-08-15
CNDIT-1585: Fix phone comments for provider work phone entry

sveselev:
## Description
Fix for `phone_comments` value for `nrt_provider` message and table when incoming data contains corresponding value.

- Phone.java: added `phone_comments` field to DTO class
- PersonExtendedProps.java: setter method for phone comments

## JIRA[CNDIT-1585](https://cdc-nbs.atlassian.net/browse/CNDIT-1585)

### DIFF
--- a/person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/PersonExtendedProps.java
+++ b/person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/PersonExtendedProps.java
@@ -26,7 +26,6 @@ public interface PersonExtendedProps {
     }
 
     default void setAliasNickname(String aliasNickname) {
-
     }
 
     void setStreetAddress1(String streetAddress1);
@@ -84,6 +83,8 @@ public interface PersonExtendedProps {
 
     default void setPhoneHome(String phoneHome) {
     }
+
+    default void setPhoneComments(String phoneComments) {}
 
     default void setPhoneExtHome(String phoneExtHome) {
     }
--- a/person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java
+++ b/person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java
@@ -23,6 +23,8 @@ public class Phone implements ExtendPerson {
     private String cd;
     @JsonProperty("ph_tl_uid")
     private Long teleLocatorUid;
+    @JsonProperty("phone_comments")
+    private String phoneComments;
 
     public <T extends PersonExtendedProps> T updatePerson(T personFull) {
         personFull.setPhElpCd(cd);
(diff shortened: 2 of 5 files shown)


person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/PersonExtendedProps.java
```diff
@@ -85,6 +84,8 @@ default void setPhoneExtWork(String phoneExtWork) {
     default void setPhoneHome(String phoneHome) {
     }
 
+    default void setPhoneComments(String phoneComments) {}
```
upasanapattnaik-eq, line 87: 👍

person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java
```diff
@@ -23,6 +23,8 @@ public class Phone implements ExtendPerson {
     private String cd;
     @JsonProperty("ph_tl_uid")
     private Long teleLocatorUid;
+    @JsonProperty("phone_comments")
+    private String phoneComments;
```
upasanapattnaik-eq, line 27: Does this add an additional field `phone_comment` to the work phone output for nrt_patient when json payload with code `WP` for cd `'PSN' ` comes in?

sveselev: The input payload for patient does not have `phone_comments` field, also, output DTO - `PatientReporting` (which extends `PersonExtendedProps`) does not have it either. So even if it happens for PSN the default interface method `setPhoneComments` will be called and won't change any actual data in `PatientReporting` (and `nrt_patient`)

sveselev: We have similar case for some other fields that exist only in ProviderReporting and not exist in `PatientReporting`, `nm_degree`, e.g. There is no `nmDegree` field in `PatientReporting`, but this is always calling:
```
public class Name implements ExtendPerson {
...
    private String nmDegree;
...
    public <T extends PersonExtendedProps> T updatePerson(T person) {
        person.setLastNm(this.lastNm);
...
        person.setNmDegree(this.nmDegree);
        return person;
    }

Comment (upasanapattnaik-eq) [reply person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java:27]: Got it! Thanks for the clarification!

**Review by upasanapattnaik-eq — APPROVED** [review]
LGTM! 👍

File: person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java [hunk 3]
```diff
@@ -23,6 +23,8 @@ public class Phone implements ExtendPerson {
     private String cd;
     @JsonProperty("ph_tl_uid")
     private Long teleLocatorUid;
+    @JsonProperty("phone_comments")
```
alog-enquizit, line 26: Optional: JsonProperty might not be required in this case since we use the Gson CamelCase library call transformer to automatically convert. We can use JsonProperty if the incoming and the transformed names are totally different.

We can still use the JsonProperty here. No Harm!

alog-enquizit (approved):
LGTM 👍

person-service/src/main/java/gov/cdc/etldatapipeline/person/model/dto/persondetail/Phone.java
sveselev: Right! Let's make it consistent.

PR #19, opened 2024-08-16
CNDIT-1586 Fix fn_get_record_status to return for INACTIVE value

sveselev:
## Description

Fixes the `fn_get_record_status` function to return `INACTIVE` value when the source column value = INACTIVE

## JIRA
[CNDIT-1586](https://cdc-nbs.atlassian.net/browse/CNDIT-1586)


upasanapattnaik-eq (commented):
LGTM 👍

upasanapattnaik-eq (approved):

upasanapattnaik-eq:
This stored procedure is also present in the rdb. That should also be updated.

PR #22, opened 2024-08-22
CNDIT-1588: Fix for updating investigation confirmation method

sveselev:
## Description
Fixes the investigation confirmation method in the investigation post-processing sp. To support multi-selected value it requires to clear the `CONFIRMATION_METHOD_GROUP` table records related to the processing investigation and replace them by new records from the `nrt_investigation_confirmation` table.

## JIRA
[CNDIT-1588](https://cdc-nbs.atlassian.net/browse/CNDIT-1588)

### DIFF
--- a/liquibase-service/src/main/resources/db/rdb/023-sp_nrt_investigation_postprocessing-001.sql
+++ b/liquibase-service/src/main/resources/db/rdb/023-sp_nrt_investigation_postprocessing-001.sql
@@ -428,7 +428,7 @@ BEGIN
 
         BEGIN TRANSACTION;
 
-        SET @proc_step_name = 'Update CONFIRMATION_METHOD and CONFIRMATION_METHOD_GROUP';
+        SET @proc_step_name = 'Update CONFIRMATION_METHOD';
         SET @proc_step_no = 3;
 
         /*Temp Confirmation Method Table*/
@@ -456,15 +456,7 @@ BEGIN
                             on cmt.confirmation_method_key = cm.confirmation_method_key
                                 and cmt.CONFIRMATION_METHOD_KEY is not null;
 
-        update cmg
-        set cmg.CONFIRMATION_DT = cmt.CONFIRMATION_DT
-        from #temp_cm_table cmt
-                 inner join dbo.confirmation_method_group cmg with (nolock)
-                            on cmt.investigation_key = cmg.investigation_key
-                                and cmt.confirmation_method_key = cmg.confirmation_method_key
-                                and cmt.CONFIRMATION_METHOD_KEY is not null;
-
-/*Logging*/
+        /*Logging*/
         SET @rowcount = @@rowcount
         INSERT INTO [dbo].[job_flow_log]
         (batch_id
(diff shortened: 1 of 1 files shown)


liquibase-service/src/main/resources/db/rdb/023-sp_nrt_investigation_postprocessing-001.sql
alog-enquizit, line 1: We will need to delete the liquibase config table row with this script entry and then run liquibase scripts

alog-enquizit (approved):
lgtm 👍

liquibase-service/src/main/resources/db/rdb/023-sp_nrt_investigation_postprocessing-001.sql
```diff
@@ -509,14 +501,44 @@ BEGIN
...
+               ,@proc_step_no
+               ,@proc_step_name
+               ,@rowcount
+               ,LEFT(@id_list, 500));
+
+        SET @proc_step_name = 'UPDATE CONFIRMATION_METHOD_GROUP';
+        SET @proc_step_no = 5;
+
+        delete dbo.CONFIRMATION_METHOD_GROUP
+        where investigation_key in
```
upasanapattnaik-eq, line 528: Are there any cases or checks needed to confirm that these are the UIDs that need to be deleted?
Eg. If we get a list of 2 updates, 2 insert uids- of which 1 insert uid matches the requirement for delete and update, are we making sure we aren't deleting all the keys that are part of the update and regular insert?

upasanapattnaik-eq: This is referencing Investigation directly

sveselev: I think there is no way to delete investigation, it can be only marked as inactive? And because sp does not have a block for investigation delete (insert/update only) we're safe here.

upasanapattnaik-eq: Ah I meant for confirmation_method_group's delete statement. It uses only Investigation table as a reference. 
Would this delete investigation keys that came in as an update uid in the @id_list?

sveselev: It will delete the records containing investigation keys related to uids from @id_list, and the immediately insert new keys (investigation and confirmation method) related to payload that came for update. We always have confirmation method values in each upcoming investigation update. If there are no values - it means, that user unselected all the codes (=deleted confirmation method). So the procedure deletes all the codes (keys) related to the investigation and does not insert anything, because this is what user wanted.

upasanapattnaik-eq: Got it! Thank you for explaining 💯  That makes sense.

PR #24, opened 2024-08-27
RTR investigation service - add PHC datamart feature flag

sveselev:
## Description

- Add a feature flag to the investigation service for PHC datamart processing. Can be enabled/disabled by by setting the `PHC_DM_ENABLE` environment variable. If the variable is not set, the default value is `false` and feature will be turned off.
- Fix investigation update sp for the investigations without notifications

### DIFF
--- a/investigation-service/src/main/resources/application.yaml
+++ b/investigation-service/src/main/resources/application.yaml
@@ -38,5 +38,7 @@ logging:
   file:
     name: InvestigationService.log
     path: ${DA_LOG_PATH:logs}
+service:
+  phc-datamart-enable: ${PHC_DM_ENABLE:false}
 server:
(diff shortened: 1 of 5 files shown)


investigation-service/src/main/resources/application.yaml
```diff
@@ -38,5 +38,7 @@ logging:
   file:
     name: InvestigationService.log
     path: ${DA_LOG_PATH:logs}
+service:
+  phc-datamart-enable: ${PHC_DM_ENABLE:false}
```
upasanapattnaik-eq, line 42: 👍

upasanapattnaik-eq (approved):
LGTM! Thank you for updating 👍

PR #25, opened 2024-08-27
CNDIT 1558: RTR increase code coverage

sveselev:
## Description
Increase code test coverage to 90% for all Data Reporting services

## JIRA
[CNDIT-1558](https://cdc-nbs.atlassian.net/browse/CNDIT-1558)


observation-service/src/main/java/gov/cdc/etldatapipeline/observation/repository/model/dto/ObservationTransformed.java
```diff
@@ -2,10 +2,8 @@
 
 import lombok.*;
 
-@NoArgsConstructor
-@Getter
 @Setter
-@ToString
+@Getter
```
alog-enquizit, line 6: If we are using both the @Getter and @Setter then consider replacing them with @Data

post-processing-service/src/main/java/gov/cdc/etldatapipeline/postprocessingservice/service/PostProcessingService.java
```diff
@@ -138,22 +138,21 @@ public void postProcessDatamart(
             JsonNode payloadNode = objectMapper.readTree(payload);
 
             Datamart dmData = objectMapper.readValue(payloadNode.get(PAYLOAD).toString(), Datamart.class);
-            final String PAYLOAD_PFX = "For payload: " + payloadNode;
             if (Objects.isNull(dmData)) {
-                throw new NoDataException(PAYLOAD_PFX + " DataMart object is null. Skipping further processing");
+                logger.info("For payload: {} DataMart object is null. Skipping further processing", payloadNode);
```
alog-enquizit, line 142: Thanks for getting this!

alog-enquizit (approved):
LGTM 👍

observation-service/src/main/java/gov/cdc/etldatapipeline/observation/repository/model/dto/ObservationTransformed.java
sveselev: > If we are using both the @getter and @Setter then consider replacing them with @DaTa

Having `@Data` annotation affects bad for test coverage, so if there is no need to have `hashCode()`, `equals()` and `toString()` it's better to avoid it.

PR #33, opened 2024-09-09
Observation model update for NRT tables

upasanapattnaik-eq:
+ Addition of new columns to nrt_observation via alter statements

### DIFF
--- a/liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
+++ b/liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
@@ -1,0 +1,30 @@
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'ctrl_cd_display_form'   AND Object_ID = Object_ID(N'nrt_observation'))
+    BEGIN
+        ALTER TABLE nrt_observation
+            ADD ctrl_cd_display_form varchar(20);
+    END;
+
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'obs_domain_cd_st_1'   AND Object_ID = Object_ID(N'nrt_observation'))
+    BEGIN
+        ALTER TABLE nrt_observation
+            ADD obs_domain_cd_st_1 varchar(20);
+
+    END;
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'processing_decision_cd'   AND Object_ID = Object_ID(N'nrt_observation'))
+    BEGIN
+        ALTER TABLE nrt_observation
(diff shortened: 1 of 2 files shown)


liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
```diff
@@ -0,0 +1,30 @@
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'ctrl_cd_display_form'   AND Object_ID = Object_ID(N'nrt_observation'))
+    BEGIN
+        ALTER TABLE nrt_observation
+            ADD ctrl_cd_display_form varchar(20);
+    END;
+
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'obs_domain_cd_st_1'   AND Object_ID = Object_ID(N'nrt_observation'))
```
alog-enquizit, line 7: For Alter statements, it should be `IF EXISTS`

liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
```diff
@@ -0,0 +1,30 @@
...
+
+    END;
+
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'cd'   AND Object_ID = Object_ID(N'nrt_observation'))
+    BEGIN
+        ALTER TABLE nrt_observation
+            ADD cd varchar(50);
+    END;
+
+IF NOT EXISTS(SELECT 1 FROM sys.columns   WHERE Name = N'shared_ind'   AND Object_ID = Object_ID(N'nrt_observation'))
```
alog-enquizit, line 26: Same for all. please remove NOT

liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
alog-enquizit: It should be like this: 
Alter the table only if the above table exists

upasanapattnaik-eq: Ah I see. I am running an IF NOT EXISTS on the column level, but on the table level, it will be IF EXISTS.

liquibase-service/src/main/resources/db/rdb/tables/017-create_nrt_observation-002.sql
upasanapattnaik-eq: Updated to check for table and add column if it does not exist

alog-enquizit (approved):
lgtm 👍

PR #35, opened 2024-09-09
CNDIT-1702: RTR performance investigate and fix for investigation service topic duplication

sveselev:
## Description
Fixes Investigation service multiple processing of the same case when relatively large volume of investigations is updated instantly.

- **Application.yaml**: set `auto-commit-enable` to **false** and `max.poll.interval.ms` to 300000 (default value);
- **InvestigationService.java**: `add consumer.commitSync()` to manually commit offsets returned on the last poll.

## JIRA
[CNDIT-1702](https://cdc-nbs.atlassian.net/browse/CNDIT-1702)

### DIFF
--- a/investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
+++ b/investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
@@ -13,6 +13,7 @@ import gov.cdc.etldatapipeline.investigation.repository.model.reporting.Investig
 import gov.cdc.etldatapipeline.investigation.util.ProcessInvestigationDataUtil;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
+import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.common.errors.SerializationException;
 import org.modelmapper.ModelMapper;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


investigation-service/src/main/java/gov/cdc/etldatapipeline/investigation/service/InvestigationService.java
```diff
@@ -78,9 +79,11 @@ public class InvestigationService {
             topics = "${spring.kafka.input.topic-name}"
     )
     public void processMessage(String message,
-                               @Header(KafkaHeaders.RECEIVED_TOPIC) String topic) {
+                               @Header(KafkaHeaders.RECEIVED_TOPIC) String topic,
+                               Consumer<?,?> consumer) {
         logger.debug(topicDebugLog, message, topic);
         processInvestigation(message);
+        consumer.commitSync();
```
upasanapattnaik-eq, line 86: 👍

investigation-service/src/main/resources/application.yaml
```diff
@@ -4,7 +4,6 @@ spring:
       topic-name: nbs_Public_health_case
     output:
       topic-name-reporting: nrt_investigation
```
upasanapattnaik-eq, line 6: Are we removing the elastic search topic for optimization?

upasanapattnaik-eq (approved):
LGTM!

investigation-service/src/main/resources/application.yaml
sveselev: I've never seen any realization for it in this version. The topic name was never used.

upasanapattnaik-eq: Hmm I believe ES outputs were originally for the patient, provider and organization search. There isn't a requirement for the Investigation afaik. It can be taken up in later sprints when the requirement arises.

PR #38, opened 2024-09-19
CNDIT-1761: Fix IDs processing order issue for RTR post-processing

sveselev:
## Description
Address and fix the inconsistency in message processing order to ensure that `Investigation` messages are processed before their corresponding `Notification` messages, preventing its out-of-sequence processing.

## JIRA
[CNDIT-1761](https://cdc-nbs.atlassian.net/browse/CNDIT-1761)


upasanapattnaik-eq (approved):
LGTM!

PR #87, opened 2024-11-20
CNDE-1863 Refactor for legacy investigations

upasanapattnaik-eq:
## Notes

Support legacy investigation and downstream legacy case datamart updates. Updates include:
- Addition of coded, text, date, and numeric views on nrt_investigation_observations. 
- Conditional update of Investigations based on given set of form codes. 
- Minor fix for empty column values that should map to NULL instead or empty. 

## JIRA

- **Related story**: [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNDE-1863?atlOrigin=eyJpIjoiNWE5OWI3ODVjYjMwNDU5NDk5NDdiZWY5Y2YzMzFjMmEiLCJwIjoiaiJ9)

## Checklist

## Types of changes

- [ ] Bugfix
- [x] New feature
- [x] Breaking change

## Testing


liquibase-service/src/main/resources/db/rdb_modern/routines/005-sp_nrt_investigation_postprocessing-003.sql
```diff
@@ -146,6 +159,196 @@ BEGIN
...
+                SELECT
+                    tnio.public_health_case_uid
+                     ,tnio.observation_id
+                     ,tnio.branch_id
+                     ,tnio.branch_type_cd
+                     ,n.cd
+                     ,n.response
+                INTO #tmp_num
+                FROM #temp_inv_obs tnio
+                         INNER JOIN dbo.v_getobsnum n ON tnio.branch_id = n.branch_id
```
johneipe, line 242: should we use nolock on the usage of the view definition or keep it defined within the view definition ?

liquibase-service/src/main/resources/db/rdb_modern/routines/005-sp_nrt_investigation_postprocessing-003.sql
```diff
@@ -51,7 +61,7 @@ BEGIN
                nrt.investigation_status        as INVESTIGATION_STATUS,
                nrt.inv_case_status             as INV_CASE_STATUS,
                nrt.case_type_cd                as CASE_TYPE,
-               nrt.txt                         as INV_COMMENTS,
+               NULLIF(nrt.txt,'')			   as INV_COMMENTS,
```
sveselev, line 64: 👍🏻

liquibase-service/src/main/resources/db/rdb_modern/routines/005-sp_nrt_investigation_postprocessing-003.sql
```diff
@@ -146,6 +159,196 @@ BEGIN
...
+                    DAYCARE_ASSOCIATION_IND = COALESCE(tmp.DAYCARE_ASSOCIATION_IND, obs.DAYCARE_ASSOCIATION_IND),
+                    DIE_FRM_THIS_ILLNESS_IND = CASE WHEN obs.DIE_FRM_THIS_ILLNESS_IND_INV145 IS NOT NULL THEN obs.DIE_FRM_THIS_ILLNESS_IND_INV145
+                                                    ELSE COALESCE(tmp.DIE_FRM_THIS_ILLNESS_IND,obs.DIE_FRM_THIS_ILLNESS_IND_PRT103, obs.DIE_FRM_THIS_ILLNESS_IND_MEA078, obs.DIE_FRM_THIS_ILLNESS_IND_RUB162)
+                        END,
+                    IMPORT_FRM_CITY = COALESCE(tmp.IMPORT_FRM_CITY, obs.IMPORT_FRM_CITY),
+                    HSPTL_ADMISSION_DT = COALESCE(tmp.HSPTL_ADMISSION_DT, obs.HSPTL_ADMISSION_DT),
+                    HSPTL_DISCHARGE_DT = COALESCE(tmp.HSPTL_DISCHARGE_DT, obs.HSPTL_DISCHARGE_DT),
+                    HSPTL_DURATION_DAYS = COALESCE(tmp.HSPTL_DURATION_DAYS, obs.HSPTL_DURATION_DAYS)
+                FROM #temp_inv_table tmp
+                         LEFT JOIN #final_inv_obs obs on tmp.case_uid = obs.public_health_case_uid;
```
joshua-mills-cdw, line 325: 👍 I was checking to see if this could be refactored into #temp_inv_table's definition, but it seems like it is better to leave this as an update statement so you can make #temp_inv_table first and you only have to filter one time and use it to make all of the subsequent temp tables. Was this also your reasoning?

liquibase-service/src/main/resources/db/rdb_modern/routines/005-sp_nrt_investigation_postprocessing-003.sql
upasanapattnaik-eq: Both can exist as well 👍 I'll update it accordingly.

liquibase-service/src/main/resources/db/rdb_modern/routines/005-sp_nrt_investigation_postprocessing-003.sql
upasanapattnaik-eq: Yes. This has me thinking of an alternative approach now- _left join_ instead of _update with COALESCE_. Functionally what this needs update the null values or leave the value as is- a left join with case statement should be able to do that as well.

joshua-mills-cdw: I think you have a pretty good case for leaving this as an update statement, but I am interested to know what's faster between the two options. Worth mentioning that even if you do make a new temp table, you can still use COALESCE instead of a case statement where applicable. Would be a little more succinct.

sveselev (approved):
LGTM!

PR #115, opened 2024-12-17
CNDE-2016: Fixed JOIN issue and refactored for performance

johneipe:
## Notes

This includes the following fixes:
- Align F_PAGE_CASE proc to use Look up tables to capture the correct Keys
- Performance improvement to both F_PAGE_CASE and F_STD_PAGE_CASE procedures

## JIRA

- **Related story**: [CNDE-2016](https://cdc-nbs.atlassian.net/browse/CNDE-2016)

## Checklist

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change

## Testing

### DIFF
--- a/liquibase-service/src/main/resources/db/rdb_modern/routines/012-sp_f_page_case_postprocessing-001.sql
+++ b/liquibase-service/src/main/resources/db/rdb_modern/routines/012-sp_f_page_case_postprocessing-001.sql
@@ -60,28 +60,12 @@ BEGIN
         WHERE inv.public_health_case_uid IN (SELECT value FROM STRING_SPLIT(@phc_ids, ','))
           AND INVESTIGATION_FORM_CD  NOT IN ( 'INV_FORM_BMDGAS','INV_FORM_BMDGBS','INV_FORM_BMDGEN',
                                               'INV_FORM_BMDNM','INV_FORM_BMDSP','INV_FORM_GEN','INV_FORM_HEPA','INV_FORM_HEPBV','INV_FORM_HEPCV',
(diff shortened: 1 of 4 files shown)


liquibase-service/src/main/resources/db/rdb_modern/routines/012-sp_f_page_case_postprocessing-001.sql
```diff
@@ -212,32 +196,37 @@ BEGIN
             drop table #DIMENSION_KEYS_PAGECASEID
             ;
 
-        select L_INV_ADMINISTRATIVE_INC.PAGE_CASE_UID as PAGE_CASE_UID
+        with LOOKUPCTE as (
```
upasanapattnaik-eq, line 199: More on a performance question- does using a cte have any impact on the time/performance for this section? Switching to the full lookup table across all the page builder dimensions would pull a lot more data as opposed to just the incremental tables.

johneipe: good point, databases usually optimize and make it work like a join even when we have CTEs. But that being said I'm not 100% sure if that is the case with SQL Server. Let me check if I can get to the query profile and see the difference.

johneipe: In the query plan, the predicate is applied (join/where condition) is applied in the single step. We also have index created on the page_case_uid columns on these tables.

PR #139, opened 2025-01-21
Cnde 2104 : Liquibase add the runOnChange tags for the db-general query

alog-enquizit:
Cnde 2104 : Liquibase add the runOnChange tags for the db-general query


upasanapattnaik-eq (approved):
LGTM!

PR #216, opened 2025-03-06
CNDE-2351,2352,2353: update to legacy case join logic, v_rdb_obs_mapping

joshua-mills-cdw:
## Notes

Two bugfixes in this PR:

- Added `batch_id` logic as necessary into the joins for the views that get data for legacy cases 
- Added always NULL rows to `v_rdb_obs_mapping` in order to always have all columns included for pivots in legacy case procedures

## JIRA

- **Related story**: [CNDE-2351](https://cdc-nbs.atlassian.net/browse/CNDE-2351)
- **Related story**: [CNDE-2352](https://cdc-nbs.atlassian.net/browse/CNDE-2352)
- **Related story**: [CNDE-2353](https://cdc-nbs.atlassian.net/browse/CNDE-2353)

## Checklist

- [ ] PR focuses on a single story.

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change

## Testing

### DIFF
--- a/db/upgrade/rdb_modern/views/001-v_getobscode.sql
+++ b/db/upgrade/rdb_modern/views/001-v_getobscode.sql
@@ -51,7 +51,10 @@ WITH InvFormQObservations AS
              FROM
                  dbo.nrt_investigation_observation tnio with (nolock)
                      inner join dbo.nrt_observation_coded ovc with (nolock) ON ovc.observation_uid = tnio.branch_id
-                     inner join dbo.nrt_observation o with (nolock) ON o.observation_uid = ovc.observation_uid
+                     inner join dbo.nrt_observation o with (nolock) ON o.observation_uid = ovc.observation_uid and 
+                     CASE WHEN ovc.batch_id IS NOT NULL and o.batch_id = ovc.batch_id then 1
+                          WHEN ovc.batch_id IS NULL then 1
+                     else 0 end = 1
              WHERE branch_type_cd = 'InvFrmQ'
 
          )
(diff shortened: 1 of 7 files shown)


db/upgrade/rdb_modern/views/001-v_getobscode.sql
```diff
@@ -51,7 +51,10 @@ WITH InvFormQObservations AS
              FROM
                  dbo.nrt_investigation_observation tnio with (nolock)
                      inner join dbo.nrt_observation_coded ovc with (nolock) ON ovc.observation_uid = tnio.branch_id
-                     inner join dbo.nrt_observation o with (nolock) ON o.observation_uid = ovc.observation_uid
+                     inner join dbo.nrt_observation o with (nolock) ON o.observation_uid = ovc.observation_uid and
```
upasanapattnaik-eq, line 54: Pretty straightforward. If batch_id exists then we get the associated data/if it is null we pull in the ids. At some point in our environment we will converge to batch ids existing for all child records so this makes sense. 
Are there any performance improvements expected over a subquery approach?

db/upgrade/rdb_modern/views/001-v_getobscode.sql
```diff
@@ -80,7 +83,7 @@ SELECT
     END AS response_cd,
     label
 FROM InvFormQObservations obs
-         LEFT JOIN dbo.codeset cs with (nolock) on cs.cd = obs.cd
+         LEFT JOIN dbo.v_codeset cs with (nolock) on cs.cd = obs.cd
```
upasanapattnaik-eq, line 86: Those higher environments need SRTE enabled 🫡

db/upgrade/rdb_modern/views/001-v_getobscode.sql
joshua-mills-cdw: As far as performance vs a subquery approach I'm not certain, but it is just the case that a subquery was not necessary for the view updates. Also, I have changed this to use ISNULL after talking with Sergei

upasanapattnaik-eq: That works!

sveselev (approved):
LGTM!

PR #354, opened 2025-04-14
CNDE-2513: Code fixes to upstream procs in the dynamic data mart process

johneipe:
## Notes

Code fixes to upstream procs in the dynamic datamart process which was identified through integrated testing

## JIRA

- **Related story**: [CNDE-2513](https://cdc-nbs.atlassian.net/browse/CNDE-2513)

## Checklist

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change

## Testing

### DIFF
--- a/db/upgrade/rdb_modern/routines/115-sp_dyn_dm_invest_form_postprocessing.sql
+++ b/db/upgrade/rdb_modern/routines/115-sp_dyn_dm_invest_form_postprocessing.sql
@@ -68,22 +68,45 @@ BEGIN
         SET @Proc_Step_Name = ' GENERATING  '+ @tmp_DynDm_INVESTIGATION_DATA;
 
 
-        DECLARE @listStr VARCHAR(MAX)
+        DECLARE @listStr VARCHAR(MAX);
+
+        SELECT @listStr = COALESCE(@listStr+',' ,'') + RDB_COLUMN_NM  + ' '+ coalesce(USER_DEFINED_COLUMN_NM,'')
+        FROM  dbo.V_NRT_NBS_INVESTIGATION_RDB_TABLE_METADATA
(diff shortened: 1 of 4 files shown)


db/upgrade/rdb_modern/routines/115-sp_dyn_dm_invest_form_postprocessing.sql
```diff
@@ -68,22 +68,45 @@ BEGIN
...
-        SET @temp_sql = '
- 		SELECT  rdb_column_nm_list , isd.INVESTIGATION_KEY
-	    into '+@tmp_DynDm_INVESTIGATION_DATA+'
-	    FROM
-			dbo.INVESTIGATION inv with ( nolock)
-	    INNER JOIN
-			#tmp_DynDm_SUMM_DATAMART isd ON	isd.INVESTIGATION_KEY  =inv.INVESTIGATION_KEY
-	    INNER JOIN
-			dbo.V_NRT_NBS_INVESTIGATION_RDB_TABLE_METADATA inv_meta on isd.DISEASE_GRP_CD =  inv_meta.INVESTIGATION_FORM_CD';
+        if len(@listStr) > 1
```
upasanapattnaik-eq, line 84: Just to round out understanding- if there are columns that exist for the form's investigation table, those should be pulled in, and there could be cases where we don't have any investigation columns as well?

johneipe: yes, I don't think we will hit that condition but i think that is how the current code functions. They have that default table getting created with just the investigation key so that downstream it doesn't fail on the join conditions.

upasanapattnaik-eq (approved):
LGTM!

PR #356, opened 2025-04-14
CNDE-2372: Fixes from CreateDM validation

joshua-mills-cdw:
## Notes

This PR includes fixes for some issues that were found during the validation of CreateDM for the dynamic datamarts.

NOTE: Indexes were not originally part of the RTR adaptation for this stored procedure, but after considering the scenario of a full load, it's best that we index on the INVESTIGATION_KEY column since many joins need to take place.

## JIRA

- **Related story**: [CNDE-2372](https://cdc-nbs.atlassian.net/browse/CNDE-2372)

## Checklist

## Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change

## Testing


liquibase-service/src/main/resources/db/rdb_modern/routines/245-sp_dyn_dm_createdm_postprocessing-001.sql
```diff
@@ -20,11 +20,61 @@ BEGIN
 
         DECLARE @datamart_suffix varchar(100) = @DATAMART_NAME+'_'+CAST(@batch_id AS varchar(50))
 
-        DECLARE @tmp_DynDm_INCOMING_DATA varchar(200) = 'dbo.tmp_DynDm_INCOMING_DATA_'+@datamart_suffix;
+        DECLARE @tmp_DynDm_INCOMING_DATA varchar(200) = 'tmp_DynDm_INCOMING_DATA_'+@datamart_suffix;
 
 
     DECLARE @tgt_table_nm NVARCHAR(200) = 'DM_INV_' + @DATAMART_NAME;
```
upasanapattnaik-eq, line 27: Do we want to create indexes on the temp tables?

johneipe: i think we can take them out and introduce later if we think it is needed.

joshua-mills-cdw: I don't think this table in particular will need an index, just the ones that are used to create it. If performance is bad during the insert/update with a large number of records then it would be easy to just add it in. Would be worth investigating whether or not it's worth it to add an index on INVESTIGATION_KEY to the target datamart tables as well, if that doesn't count as modifying the existing RDB schema.

PR #690, opened 2026-03-17
APP-299: Resolve NULL for D_ORGANIZATION in RTR

nullflux:
This fix combines both Eli and Michael's approaches in order to cover both a missing entry in CVG _and_ an empty string in CVG. Just the `COALESCE` would likely work, but a little extra caution couldn't hurt.

## JIRA

- **Related story**: [APP-299](https://cdc-nbs.atlassian.net/browse/APP-299)

## Checklist

## Types of changes

- [x] Bugfix

## Testing

I manually reproduced this bug by creating an Organization in the UI and examining the result in RDB_MODERN. Then, I updated the stored procedure with this fix, and verified that creating a new Organization resulted in this column being populated.

Side note: of the various types of "Identification," in the Organization edit flow, this _must_ be a "Facility Id." Per the CASE statement in the line above where I've added this fix, no other ID type will be populated into these columns.


mpeels (approved):

hawley-skylight (approved):
thanks!

PR #693, opened 2026-03-18
APP-382: OBJECT_ID doesn't allow # temp table shorthand

nullflux:
## Notes

From the [TSQL docs](https://learn.microsoft.com/en-us/sql/t-sql/functions/object-id-transact-sql?view=sql-server-ver17#remarks):

> When a temporary table name is specified, the database name must come before the temporary table name, unless the current database is tempdb. For example:
> 
> `SELECT OBJECT_ID('tempdb..#mytemptable');`

## JIRA

- **Related story**: [APP-382](https://cdc-nbs.atlassian.net/browse/APP-382)

## Checklist

## Types of changes

- [x] Bugfix

## Testing

I have verified that the migrations that install the stored procedures still _run_ and result in the procedures being installed but have otherwise not done any functional testing around this change.


mehansen (approved):
🚢

ericbuckley (approved):

PR #713, opened 2026-03-25
Fix topic name in Observation Service

mehansen:
## Description
The observation service is failing to start with the following error:
```
Could not resolve placeholder 'spring.kafka.input.topic-name-person' in value "${spring.kafka.input.topic-name-person}"
```
It looks like the topic name placeholder was accidentally changed in #696.

## Related Issue
#696 

## Checklist


mpeels (approved):
I believe this was unintentional as the other PR should have only adjusted it within the context of the hydration service.

adaball (approved):
Apologies, that change was not planned in my original PR but it seems I inadvertently altered it.  

We combined a number of configs into the config for `reporting-hydration-service` which included altering a number of topic config keys and it's certainly possible I opened the original `ObservationService` in my IDE instead of the consolidated version.

PR #749, opened 2026-04-09
fix reseeding new nrt_lab_test_key table

mehansen:
## Description
The reseed step after creating this table technically shouldn't matter because it's reseeded again during onboarding here:
https://github.com/CDCgov/NEDSS-DataReporting/blob/2b3b02e20f44e3d400e3c0970c30408b20b6ee8b/liquibase-service/src/main/resources/db/001-master/02_onboarding_script_data_load/999-reseed_nrt_key-001.sql#L470-L481
However it feels like good practice to never have it be seeded in a way that will conflict with the keys in LAB_TEST.

## Related Issue
discovered while looking into [APP-387](https://cdc-nbs.atlassian.net/browse/APP-387)

## Additional Notes
From the [docs](https://learn.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-checkident-transact-sql?view=sql-server-ver17#remarks) for CHECKIDENT:
> Current identity value is set to the new_reseed_value. If no rows have been inserted into the table since the table was created, or if all rows have been removed by using the TRUNCATE TABLE statement, the first row inserted after you run DBCC CHECKIDENT uses new_reseed_value as the identity. If rows are present in the table, or if all rows have been removed by using the DELETE statement, the next row inserted uses new_reseed_value + the [current increment](https://learn.microsoft.com/en-us/sql/t-sql/functions/ident-incr-transact-sql?view=sql-server-ver17) value.

So basically the value of the next row inserted after a reseed depends on whether the table has no rows or not. It's safe to reseed to `@max` except when the table is empty 🥴 

## Checklist


liquibase-service/src/main/resources/db/005-rdb_modern/tables/248-create_nrt_lab_test_key-001.sql
```diff
@@ -8,12 +8,14 @@ IF NOT EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_lab_test_key' and xtyp
...
 		SELECT @max=max(LAB_TEST_KEY) from [dbo].LAB_TEST;
 		SELECT @max;
-		IF @max IS NULL   --check when max is returned as null
-			SET @max = 1; -- default to 1, default record with key = 1 is already created
-		DBCC CHECKIDENT ('[dbo].nrt_lab_test_key', RESEED, @max);
+		IF @max IS NULL
+			SET @max = 1; -- LAB_TEST is expected to have a default null row with LAB_TEST_KEY = 1
+		-- when a table with no rows is reseeded, the third parameter will be the key value of the first added row, so
+        -- we must increment by 1 to prevent key collision with LAB_TEST
+		DBCC CHECKIDENT ('[dbo].nrt_lab_test_key', RESEED, @max+1);
```
mpeels: Dang off by one errors. Good catch!

liquibase-service/src/main/resources/db/001-master/02_onboarding_script_data_load/0999-reseed_nrt_key-001.sql
```diff
@@ -85,16 +85,6 @@ IF EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_notification_key' and xtyp
...
-IF EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_lab_test_key' and xtype = 'U')
-	BEGIN
-		
-		SELECT @max=max(LAB_TEST_KEY) from [dbo].LAB_TEST;
-		SELECT @max;
-		IF @max IS NULL   --check when max is returned as null
-			SET @max = 2; -- default to 2, default record with key = 1 is already created
-		DBCC CHECKIDENT ('[dbo].nrt_lab_test_key', RESEED, @max);
-
-	END
```
mehansen, line 97: this block of code is in this file twice so I removed the first appearance

mpeels (approved):

ericbuckley:
@mehansen [there are a number of tables](https://github.com/search?q=repo%3ACDCgov%2FNEDSS-DataReporting%20%20%20%20%20%20%20%20%22RESEED%2C%20%40max%29%3B%22&type=code) in rdb_modern that have entries like this shortly after the tables are created, do we need to make changes to all of these tables?

`﻿DBCC CHECKIDENT ('*', RESEED, @max);`

mehansen:
> @mehansen [there are a number of tables](https://github.com/search?q=repo%3ACDCgov%2FNEDSS-DataReporting%20%20%20%20%20%20%20%20%22RESEED%2C%20%40max%29%3B%22&type=code) in rdb_modern that have entries like this shortly after the tables are created, do we need to make changes to all of these tables?
> 
> `﻿DBCC CHECKIDENT ('*', RESEED, @max);`

@ericbuckley good point! it seems like many of them are doing the addition in the line when they assign the `@max` value, but I see at least 1 other that looks wrong. I'll put this back in draft until I check them out

mehansen:
after reviewing the usages of `DBCC CHECKIDENT`, I found a couple instances where we are seeding with 1 and definitely shouldn't because I could find an explicit place in the code where we filter out that key = 1 (e.g. [organization_key](https://github.com/CDCgov/NEDSS-DataReporting/blob/4db7a5ff1819be81304acb1843194de95d5a65a6/liquibase-service/src/main/resources/db/005-rdb_modern/routines/017-sp_d_labtest_result_postprocessing-001.sql#L1122-L1123)) and I also found some where it wasn't so explicit (e.g. case_management_key) but it seemed safer just to always skip 1 when seeding since there's no requirement to have a row where key = 1
also rewrote my previous change to match the established style for seeding tables

liquibase-service/src/main/resources/db/001-master/02_onboarding_script_data_load/0999-reseed_nrt_key-001.sql
ericbuckley: I noticed in that block, lines 480 to 490, the code is not incrementing the `@max` value by 1 if entries are found in LAB_TEST.  Do we also need to correct those statements?

For example
`SELECT @max=max(LAB_TEST_KEY) from [dbo].LAB_TEST;`
vs
`SELECT @max=max(LAB_TEST_KEY)+1 from [dbo].LAB_TEST;`

mehansen: I don't think so. at the point this reseed script is running, `001-nrt_lab_test_key-001.sql` will have already ran, which will populate any keys from `LAB_TEST` into `nrt_lab_test_key`. so if `LAB_TEST` is not empty, the table will have rows in it by now and then the reseed behavior is "new_reseed_value + increment value"
>If rows are present in the table, or if all rows have been removed by using the DELETE statement, the next row inserted uses new_reseed_value + the [current increment](https://learn.microsoft.com/en-us/sql/t-sql/functions/ident-incr-transact-sql?view=sql-server-ver17) value.

I wonder if most if not all of the reseeds done in this file could skip the +1. I'm guessing it was done out of an abundance of caution or because the behavior for this command is not that easy to follow

PR #755, opened 2026-04-20
Direct insert of nrt_observation_* data

mpeels:
## Description
In order to resolve the race condition between the PostProcessingService and the Kafka Sink Connector for observation data, the ObservationService code was updated to directly persist to the following tables:
<ul>
  <li>nrt_observation_material
  <li>nrt_observation_coded
  <li>nrt_observation_date
  <li>nrt_observation_edx
  <li>nrt_observation_numeric
  <li>nrt_observation_reason
  <li>nrt_observation_txt
</ul>

## Related Issue
[APP-519](https://cdc-nbs.atlassian.net/browse/APP-519)

## Additional Notes
1. The `ObservationService.java` class was previously responsible for listening for Kafka messages as well as some of the business logic for processing the messages. It has been renamed to `ObservationConsumer.java` and updated to only validate messages and extract `Organization_uid` values. The new `ObservationProcessor.java` class now handles the business logic of processing the data.
2. The `NrtObservationWriter` is using raw sql upsert statements, with a few more hours work, JPA Repositories could be created with `@Entity`'s defined for each table to handle the insert/update if desired.
3. The previous code had a single class, `ProcessObservationDataUtil.java`, that was responsible for both parsing the Observation data and posting messages to the relevant Kafka Topics. This was split into an `ObservationParser.java` and the `NrtObservationWriter.java`.
4. Resolves a possible thread safety issue where the [ObservationKey](https://github.com/CDCgov/NEDSS-DataReporting/blob/d8b378e4441b0fce548c009cf89bf7e1fe8e9eba/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/observation/service/ObservationService.java#L93) was declared at the class level but multiple [threads](https://github.com/CDCgov/NEDSS-DataReporting/blob/d8b378e4441b0fce548c009cf89bf7e1fe8e9eba/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/observation/service/ObservationService.java#L146-L147)  were [setting its value](https://github.com/CDCgov/NEDSS-DataReporting/blob/d8b378e4441b0fce548c009cf89bf7e1fe8e9eba/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/observation/service/ObservationService.java#L173) simultaneously.

## Checklist


reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/observation/repository/NrtObservationWriter.java
ericnagel, line 1: In your updates, you do not need to update the fields you joined on, as you know those values are not changing. You skipped this in `nrt_observation_material` but not the others. Either add this into `nrt_observation_material` (unnecessary, but it's nice having all fields explicitly listed), or remove the following lines (data is not changing, so don't set them; work from the bottom up, so the line numbers stay in tact):
- 164
- 165
- 232
- 278
- 279
- 323
- 390
- 391
- 434
- 435

mpeels:
Will re-open once we revisit these race conditions

PR #758, opened 2026-04-20
APP-515 rdb date and condition setup

hawley-skylight:
## Description
This ensures that the RDB_Modern..Condition is populated on install from the information in the already
populated RDB_Modern..NRT_SRTE tables. 

This also creates the RDB_MODERN..RDB_DATE table if it does not exist already, and adds a standard null row to it.

This is sufficient to unblock the functional testing so that test03 runs without error, and the relevant
datamarts NOTIFICATION, NOTIFICATION_EVENT, GENERIC_CASE, LDF_FOODBORNE receive data.

Not in this pull request is code to pre-populate RDB_DATE, as it is unclear what the correct algorithm
is to fill the table. We still need to confirm what the correct data is. If we use dev71 as a reference, 
RDB_DATE appears to be populated according to the MMWR rules, but it is not sure whether or not
that data in dev71 is correct as a model.

## Related Issue
[APP-515](https://cdc-nbs.atlassian.net/browse/APP-515)

## Additional Notes

## Checklist
- [x ] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [ x] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [ x] I have updated the documentation, if applicable.
- [ x] I have added or updated test cases to cover my changes, if applicable.


liquibase-service/src/main/resources/db/005-rdb_modern/tables/258-create_rdb_date-001.sql
```diff
@@ -0,0 +1,25 @@
+-- Create RDB_DATE reference table if it does not yet exist.
+
+IF NOT EXISTS (SELECT 1 FROM sysobjects WHERE name = 'RDB_DATE' AND xtype = 'U')
```
ericbuckley, line 3: Doesn't this migration already create the schema?

https://github.com/CDCgov/NEDSS-DataReporting/blob/main/liquibase-service/src/main/resources/db/005-rdb_modern/routines/014-sp_get_date_dim-001.sql

hawley-skylight: That creates a stored procedure which, as far as I can tell, is unused.

ericbuckley: True, but if we create a new migration that executes that stored procedure with the parameters 1990 and 2050, it should create the RDB_DATE table and populate it.

ericbuckley: Actually I'm wrong about that, the proc won't create the table.  However, RDB_DATE exists in RDB, so it should be copied over when we [restore](https://github.com/CDCgov/NEDSS-DataReporting/blob/main/containers/db/initialize/01-restore-modern.sql).  Is there a situation you're running into where we don't want to build rdb_modern from a backup of rdb?

ericnagel (approved):

liquibase-service/src/main/resources/db/005-rdb_modern/tables/258-create_rdb_date-001.sql
hawley-skylight: Correct, when I tried it, it didn't create the needed temp tables. It also sometimes had errors when running due to foreign key constraints from other tables to rdb_date, depending on the current state when the sp is run. So I tried a simpler .sql to populate the table idempotently without using a temp table and found that the algorithm used by sp_get_date_dim doesn't quite match the contents of dev71 rdb_date. 

The current snapshot we are using ends up with an empty rdb_date in rdb_modern locally. We can either
change that or add a migration to fill rdb_date. Either way, we need to know what the correct values
should be in rdb_date. Perhaps we merge this one to unblock testing-- the existence of a non-empty table is sufficient for that-- and I can do further investigation about what the correct values should be in rdb_date.

ericbuckley: If this is blocking, I'm fine with merging.

Per the wrong output from the `sp_get_date_dim` function, I'd like to get that into a bug ticket.  I'm guessing the RDB.dbo.RDB_DATE contents on dev71 is correct, as that is currently being populated with the old SAS program, so the desired output should match that.  Is that your thought?

hawley-skylight: That's what I was thinking. But I'm not sure.

PR #770, opened 2026-04-28
covid case datamart fixes

ericbuckley:
## Description
Fixing a bug that prevents covid case datamart from running cleanly and converting *some* columns in those tables to display codes rather than descriptions.

## Related Issue
[APP-460](https://cdc-nbs.atlassian.net/browse/APP-460)

## Additional Notes
- Removing the use of TestContainers to orchestrate the database and liquibase migrations (this was causing timing issues with the newly added test case of data not being available at assertion time).
- Leveraging the Java Liquibase SDK to apply migrations before the stored proc unit tests begin.
- Converted as many descriptions to codes, per alignment with the Covid19ETL process, that was possible based on the data available in the RDB_Modern.   
- Updated `004-nrt_lab_test_result_group_key-001.sql` onboarding script to fix a potential primary key violation that could occur during database updates or re-onboarding by strictly enforcing uniqueness based on the table's primary identity column (this allows us to re-run the migration on the database without failure).
- Updated `DataDrivenUnitTests.java` to use a transaction block to make the tests more robust by guaranteeing a clean database state for every iteration.  This allows us to reuse the same primary keys from test to test.
- Added a `dotenv` plugin to automatically load variables from devs `.env` file into the application-test.yaml.  This allows developers to override testing settings before running gradle tests locally.
- Updated the `nrt_patient` table to include the `marital_status_cd` column, as this needed to be in the `covid_case_datamart` table
- Changed both the UnitTest and FunctionalTest abstract classes to use an runtime variable to optionally add an override docker compose file.

## Checklist


eliSkylight (approved):

mpeels:
Github will not allow me to add a comment to the code -- but regarding the removal of the TestContainers usage for the UnitTest.java:

With the complete removal of this, it is now `required` to have a database running locally prior to kicking off the unit tests and a database must `not` be running prior to starting the functional tests. 

Thoughts on having an environment variable to control wether or not to use TestContainers for the database?

ericbuckley:
> Github will not allow me to add a comment to the code -- but regarding the removal of the TestContainers usage for the UnitTest.java:
> 
> With the complete removal of this, it is now `required` to have a database running locally prior to kicking off the unit tests and a database must `not` be running prior to starting the functional tests.
> 
> Thoughts on having an environment variable to control wether or not to use TestContainers for the database?

I didn't fully think through how difference between functional and unit setup will impact development workflows.  Let me sit on that for a bit, and experiment with bringing just the nbs-mssql container back in a TestContainer

liquibase-service/src/main/resources/db/005-rdb_modern/routines/310-sp_covid_case_datamart_postprocessing-001.sql
```diff
@@ -257,41 +257,39 @@ BEGIN TRY
...
         pat.PATIENT_DOB AS 'PATIENT_DOB',
         cast(pat.PATIENT_AGE_REPORTED as varchar(10)) AS 'PATIENT_AGE_REPORTED',
-        pat.PATIENT_AGE_REPORTED_UNIT AS 'PATIENT_AGE_RPTD_UNIT',
+        nrtPat.AGE_REPORTED_UNIT_CD AS 'PATIENT_AGE_RPTD_UNIT',
         cast(pat.PATIENT_BIRTH_COUNTRY as varchar(20)) AS 'PATIENT_BIRTH_COUNTRY',
         cast(pat.PATIENT_CURRENT_SEX as varchar(1)) AS 'PATIENT_CURRENT_SEX',
-        cast(pat.PATIENT_DECEASED_INDICATOR as varchar(20)) AS 'PATIENT_DECEASED_IND',
+        cast(nrtPat.DECEASED_IND_CD as varchar(20)) AS 'PATIENT_DECEASED_IND',
         pat.PATIENT_DECEASED_DATE AS 'PATIENT_DECEASED_DT',
         pat.PATIENT_MARITAL_STATUS AS 'PATIENT_MARITAL_STS',
```
ericbuckley: @CDCgov/nbs-dragon When the covid19 ETL runs in NBS6, `PATIENT_MARITAL_STATUS` is populated with a code and not a description (eg 'M' vs 'Married').

From what I can tell of the postprocessing stored procs, there seems to be a deliberate decision to only pull data from other RDB_MODERN tables (eg nrt_* and d_*).  If we stick with that design decision, we can't change this column to display the code, as it doesn't exist in D_PATIENT or NRT_PATIENT (both of those tables only have the full marital status value, eg 'Married').

For discrepancies like these, should we be changing the person service and `sp_patient_event` stored proc to capture new data to make it possible?

mpeels: Might need to bring it up with Kasey. If the answer is "the table needs to show the code, make it happen", then I do think that modifying the upstream `sp_patient_event` to pull the data makes the most sense.

eliSkylight (approved):

eliSkylight (commented):
Sorry, should not have selected Approve due to failing checks

reporting-pipeline-service/src/test/resources/application-test.yaml
```diff
@@ -24,3 +25,7 @@ spring:
       value-deserializer: org.springframework.kafka.support.serializer.JsonDeserializer
       maxPollIntervalMs: 3000
       metadata.max.age.ms: 6000
+  liquibase:
+    enabled: false
+  testcontainers:
+    customComposeFile: ${CUSTOM_COMPOSE_FILE:}
```
ericbuckley, line 31: @mpeels I added a new variable here, one can add `CUSTOM_COMPOSE_FILE="docker-compose.test.yaml"` to their `.env` file if they want to use that to speed up the development process

eliSkylight (approved):

PR #776, opened 2026-04-29
add functional test02: mark negative Covid case reviewed

hawley-skylight:
## Description
Add functional test 02: mark a negative Covid Lab Report as reviewed.

## Related Issue
[APP-285](https://cdc-nbs.atlassian.net/browse/APP-285)

## Additional Notes

## Checklist


reporting-pipeline-service/src/test/resources/testData/functional/covidMarkReviewed/010-createPatientAndLabReport/expected.json
```diff
@@ -0,0 +1,373 @@
...
+      "LDF_GROUP_KEY": 1,
+      "MORB_RPT_KEY": 1,
+      "ORDERING_ORG_KEY": 1,
+      "ORDERING_PROVIDER_KEY": 3,
+      "PERFORMING_LAB_KEY": 1,
+      "RECORD_STATUS_CD": "ACTIVE",
+      "REPORTING_LAB_KEY": 3,
+      "RESULT_COMMENT_GRP_KEY": 1,
+      "SPECIMEN_COLLECTOR_KEY": 1,
+      "TEST_RESULT_GRP_KEY": 1
```
mpeels, line 324: I believe the `*_KEY` values might change based on future tests and cause a failure. For example, if a different test that creates a `LAB_TEST_KEY` entry runs prior to this test, the value `2` will no longer be correct here.

hawley-skylight: All but CONDITION_KEY? CONDITION_KEY is unlikely to be changed by other tests.

mpeels: Yea, i believe that `CONDITION_KEY` would be fairly static.

hawley-skylight: I updated the queries to use UIDs in WHERE, and ORDER BY a UID instead of a _KEY. That should make the tests less context dependent. However, I didn't remove _KEY from the queries because there are a number of KEYS that are stable (such as PERFORMING_LAB_KEY, CONDITION_KEY, REPORTING_LAB_KEY) given the seed data. Since the test passes as it is, I think it is simpler to leave the KEYs in, and only remove a _KEY from a query if it later causes this test to fail when this test is run after or in parallel with some new test that is added.

mpeels (approved):

PR #785, opened 2026-04-30
Align provider phone mapping with legacy behavior and exclude inactive cell phones

mehansen:
## Description
This PR addresses two JIRAs related to person phone numbers in RDB.

**APP-461 — Phone mapping mismatches in COVID Lab Datamart, `D_Provider`, and `D_Patient`**
RTR was not matching MasterETL's logic for filtering person phone numbers by `use_cd` and `cd` in several different scenarios, causing diffs in the aforementioned tables under certain circumstances. The root issue was that RTR conflated distinct phone types rather than treating them separately as MasterETL does.
The diffs were:
* `D_Provider` work phone is explicitly "office" type (cd = O, use_cd = WP) in MasterETL, but RTR filtering logic was allowing cell phones.
* `COVID_LAB_DATAMART` provider work phone is explicitly "workplace phone" type (cd = PH, use_cd = WP) in MasterETL, but RTR was referencing `nrt_provider`/`D_Provider` work phone, so getting the office phone or cell phone.
* `D_Patient` work phone is explicitly "workplace phone" type (cd = PH, use_cd = WP) in MasterETL, but RTR was allowing other `cd` values.
* `D_Patient` home phone is explicitly "home phone" type (cd = PH, use_cd = H) in MasterETL, but RTR was allowing other `cd` values.

To fix:
* Two new columns (`phone_work_phone`, `phone_ext_work_phone`) are added to `nrt_provider` to store the `PH`-type work number separately from the existing `phone_work`/`phone_ext_work` columns (which hold the `O`-type office number).
* `sp_covid_lab_datamart_postprocessing` is updated to read from the new columns for ordering provider phone. 
* The telephone query in `sp_provider_event` is simplified from a `UNION ALL` and updated to allow other `cd` values besides `O` and `CP`.
* Phone processing logic in `DataPostProcessor` and `Phone.updatePerson()` is refactored to map each phone type to the correct field based on both `useCd` and `cd`. Note that provider work phone and patient work phone are mapped differently which matches MasterETL.

**APP-595 — Inactive provider cell phone numbers incorrectly included**
Discovered during APP-461 testing: RTR was populating `D_PROVIDER.PROVIDER_PHONE_CELL` with cell phone records that had been soft-deleted in ODSE (i.e., `RECORD_STATUS_CD != 'ACTIVE'`). This was carried over from MasterETL, but SME review confirmed it is a bug. This PR fixes it by adding an `ACTIVE` record status filter to cell phone lookups for providers, consistent with how inactive records are already excluded for provider work phones and patient cell phones.

## Related Issue
[APP-461](https://cdc-nbs.atlassian.net/browse/APP-461) | [APP-595](https://cdc-nbs.atlassian.net/browse/APP-595)

## Additional Notes
Per APP-595, the fix to exclude inactive cell phones is a deliberate divergence from MasterETL behavior; will update https://cdc-nbs.atlassian.net/wiki/x/A4BYfw to document the change.

## Checklist


liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
```diff
@@ -28,8 +28,10 @@ CREATE TABLE dbo.nrt_provider
     address_comments               varchar(2000)                                null,
     phone_work                     varchar(50)                                  null,
     phone_ext_work                 varchar(50)                                  null,
-    email_work                     varchar(50)                                  null,
     phone_comments                 varchar(2000)                                null,
+    phone_work_phone               varchar(50)                                  null,
+    phone_ext_work_phone           varchar(50)                                  null,
```
mehansen, line 33: would love thoughts on better names for these, or other ways to differentiate between the two types of work phones

mpeels (approved):
Excellent

liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
ericnagel: Are these two different types of work phones, or are they work phone (ex: 716-555-1212) and work phone extension (ex: 12345)?

liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
```diff
@@ -28,8 +28,10 @@ CREATE TABLE dbo.nrt_provider
...
+    email_work                     varchar(50)                                  null,
```
ericnagel, line 34: Is 50 characters long enough for an email address?

liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
mehansen: if you look at the fields above these you'll see
* `phone_work` and `phone_ext_work` which represent the phone (and extension) entered with Use = Primary Workplace and Type = Office in the UI
* these are different from `phone_work_phone` and `phone_ext_work_phone` which represent the phone (and extension) entered with Use = Primary Workplace and Type = Phone

we need both types in the nrt_provider table to replicate MasterETL behavior, but I'm cringing at the name for the ones I added

liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
mehansen: hmm maybe not. it matches the `D_PROVIDER.PROVIDER_EMAIL_WORK` column though. we could update it across RDB but I would push for that being a separate ticket since I'm not doing anything to that field, I just changed the order so the work phone fields could be grouped together. specifically, it bothered me that `phone_comments` was separate when that field corresponds to the same phone used for `phone_work` and `phone_ext_work`

ericnagel: oh I see - then leave this alone. Github made it look like email was being added.

liquibase-service/src/main/resources/db/005-rdb_modern/tables/003-create_nrt_provider-001.sql
ericnagel: Oh then maybe:
- `phone_primary_office`
- `phone_primary_office_ext`

Also, maybe another issue but it appears that the area code is being saved in the `country_code` field.

mehansen: what makes you say that about `country_code`? are you referring to country_code = 840? because if so 840 is the ISO standard code for the United States

ericnagel: OK, I was confused because I put "1" in the country code, and one of the phone numbers had 840 as the area code. So that part is OK

mehansen: I think `phone_primary_office` / `phone_primary_office_ext` would be good names for the existing fields which represent Type = Office in the UI, but I was reluctant to rename those since the field is shared with PatientReporting via the PersonExtendedProps interface, and in the patient case they don't represent an office phone. there are ways around this, it just started feeling more complicated than it needed to be 😄 

I was wondering if there was a better name for the new columns but as I think about it I think the `_phone` postfix plus the comment on ProviderReporting are probably good enough

PR #794, opened 2026-05-05
move testing tools to tools folder

mpeels:
## Description
Moves our support tooling into a `testing-tools` folder.

Folders moved:
1. functional-testing (contains browser recordings for test workflows)
2. local-db-tracing (contains a python tool for capturing database changes)
3. performance-testing (contains ELR generator)


hawley-skylight:
if we must, how about "testing-tools"?

generally I'd advocate only increasing file hierarchy when necessary. it may feel tidy, but it
it makes it harder to find things later, given how human cognition works.
see, e.g. https://blog.ploeh.dk/2023/05/29/favour-flat-code-file-folders/

ericnagel:
You have `test05` and `test4` - I think we should:

1. Stick with padding numbers with 0's, ideally to 3 digits
2. Add the numbering to the beginning (ex: `005-test`)
3. Actually name the tests (ex: `005-long-path-supervisor-review-twice`)

ericnagel:
The local-db-testing documentation will have to be updated to reflect the new file locations, after this change is merged in.

mpeels:
> if we must, how about "testing-tools"?
> 
> generally I'd advocate only increasing file hierarchy when necessary. it may feel tidy, but it it makes it harder to find things later, given how human cognition works. see, e.g. https://blog.ploeh.dk/2023/05/29/favour-flat-code-file-folders/

We definitely don't have to make this move, its just a personal preference. I wanted an easy way to show "code we ship" vs "stuff that supports the code we ship".

ericnagel (approved):

PR #795, opened 2026-05-05
Move common-util into reporting-pipeline-service

mpeels:
## Description
Moves the `common-util` code into the reporting-pipeline-service. Theres no need to have a shared library when its a single service...

## Related Issue
🤠

## Checklist


common-util/build.gradle
```diff
@@ -1,67 +0,0 @@
...
-}
-
-dependencies {
-    compileOnly 'org.springframework.boot:spring-boot-autoconfigure'
-    compileOnly 'org.springframework.boot:spring-boot-actuator-autoconfigure'
-    api 'io.micrometer:micrometer-core'
-
-    compileOnly 'org.projectlombok:lombok'
-    annotationProcessor 'org.projectlombok:lombok'
-    compileOnly 'org.slf4j:slf4j-api'
```
ericbuckley, line 43: Do we need to merge this dep into the reporting-pipeline-service build.gradle?

ericbuckley (approved):
Looks good, just some questions on whether we need to merge plugins/deps specified in the common-util subproject into the reporting-pipeline-service subproject

common-util/build.gradle
eliSkylight: We might need to add the `compileOnly 'org.slf4j:slf4j-api'` to the pipeline service [build](https://github.com/CDCgov/NEDSS-DataReporting/pull/795/changes#diff-75bb2f66c344806d76a28ad0c788b36d8f084337d617c26c010a96c732b81623). It already has the compile entry for `lombok` and Michael added `common-util`, but wondering if `slf4j` needs to be there as well.

mpeels: Went ahead and added the `compileOnly 'org.slf4j:slf4j-api'` entry just in case.

eliSkylight (approved):

PR #816, opened 2026-05-13
Improvements to functional testing tools

ericnagel:
## Description
Handful of changes that improve the testing process (in particular, RDB vs. RDB_Modern and vice versa) including:
- Add logic to handle midnight time in datetime replacements
- Adding tables to ignore as per https://cdc-nbs.atlassian.net/wiki/spaces/NE/pages/2136506371/RTR+reporting+differences#RDB-Tables-Not-Considered-for-Comparison
- Refactor exclusion logic for artifact tables to ensure case-insensitivity
- When validating data, if the difference is a date or datetime and both sides have a value, treat it as a warning - not an error. If either side is missing a value (or is NULL), it's an error. Also, if it's a datetime and one side is midnight but the other isn't, it's an error.

## Related Issue
[APP-280](https://cdc-nbs.atlassian.net/browse/APP-280)

## Checklist

### DIFF
--- a/testing-tools/local-db-tracing/replay_setup.py
+++ b/testing-tools/local-db-tracing/replay_setup.py
@@ -189,25 +189,26 @@ def value_has_midnight_time(value_token: str) -> bool:
 
 
 def replacement_expression(sql_type: str, has_midnight_time: bool = False) -> str:
+    sql_current_date = "CAST(CURRENT_TIMESTAMP AS date)"
     lowered = sql_type.lower()
     if lowered == "date":
-        return "CURRENT_DATE"
+        return sql_current_date
     if lowered == "time":
         return "CAST(CURRENT_TIMESTAMP AS time)"
     if lowered.startswith("datetimeoffset"):
         if has_midnight_time:
-            return "CAST(CURRENT_DATE AS datetimeoffset)"
+            return f"CAST({sql_current_date} AS datetimeoffset)"
         return "CAST(CURRENT_TIMESTAMP AS datetimeoffset)"
     if lowered.startswith("datetime2"):
         if has_midnight_time:
-            return "CAST(CURRENT_DATE AS datetime2)"
+            return f"CAST({sql_current_date} AS datetime2)"
         return "CAST(CURRENT_TIMESTAMP AS datetime2)"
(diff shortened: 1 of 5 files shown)


testing-tools/local-db-tracing/replay_setup.py
```diff
@@ -162,25 +162,78 @@ def normalize_column_name(column_token: str) -> str:
     return stripped
 
 
-def replacement_expression(sql_type: str) -> str:
+def value_has_midnight_time(value_token: str) -> bool:
+    """Check if a date/datetime literal has a midnight time component (00:00:00)."""
+    stripped = value_token.strip()
```
ericbuckley, line 167: I'm not sure if you opted for the string comparisons for performance reasons, but if not, you could always use `datetime.strptime` to convert it to a DT object first, then make your checks.

testing-tools/local-db-tracing/validate_rdb_selects.py
```diff
@@ -340,6 +377,33 @@ def normalize_trailing_zero_millis(value: object) -> object:
     return re.sub(r"(?<=\d)\.000(?=(?:Z|[+-]\d{2}:?\d{2})?$)", "", value)
 
 
+def is_iso_datetime_or_date_string(value: object) -> bool:
```
ericbuckley, line 381: There is a built-in that already does this and it checks for valid dates too (not sure if that matters, eg "2026-20-40").  https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

ericbuckley (approved):

hawley-skylight (approved):

testing-tools/local-db-tracing/replay_setup.py
ericnagel: Yeah, we're not doing anything more than checking if it's midnight or not so keeping it a string is (slightly) quicker. If we were doing more with the date (or time), using a DT object makes sense.

testing-tools/local-db-tracing/validate_rdb_selects.py
ericnagel: Updated `is_iso_datetime_or_date_string` to use `datetime.fromisoformat`

PR #826, opened 2026-05-18
Fix sp_nrt_provider_postprocessing: correct #PROVIDER_UPDATE_LIST typo

nullflux:
## Description
`sp_nrt_provider_postprocessing` at line 564 referenced `#PATIENT_UPDATE_LIST`, but the only temp table in scope at that point is `#PROVIDER_UPDATE_LIST`. The UPDATE-with-diff path therefore failed with Msg 208 "Invalid object name '#PATIENT_UPDATE_LIST'" whenever an existing provider had a diff against an incoming `nrt_provider` row.

This PR fixes the single-character typo. Latent on a baseline DB (the UPDATE-with-diff path only fires when an existing `D_PROVIDER` row gets a diff), but trips deterministically on the first such case.

Verified locally: re-applying the routine and re-running the repro causes the SP to complete the diff-update path cleanly (pre-fix it raised Msg 208; post-fix `job_flow_log` shows no error rows).

## Related Issue

## Additional Notes
A `testData/unit` fixture is included that seeds an existing `D_PROVIDER` row and an `nrt_provider` row with field-level differences, then asserts `D_PROVIDER` is updated to reflect the new values. Note: `nrt_provider` is a system-versioned temporal table; the fixture excludes the `GENERATED ALWAYS` columns from its INSERT lists.

## Checklist


eliSkylight (approved):

eliSkylight:
Code looks good, but maybe the tests need to run again?
```
DataDrivenUnitTests > testRunner(Path) > [3] testDirectory=src/test/resources/testData/unit/bug4_provider_update_typo FAILED
    org.springframework.dao.CannotAcquireLockException at DataDrivenUnitTests.java:102
        Caused by: com.microsoft.sqlserver.jdbc.SQLServerException at DataDrivenUnitTests.java:102
```

mpeels (approved):

ericbuckley (approved):

mehansen:
~I think this will fix the provider part of the datamart sync, which currently we're unsure if we even want to keep implemented. I think this bug was preventing it from ever firing `sp_provider_dim_columns_update_to_datamart`. is there a chance we don't want to fix this and instead should rip it out if it's already not working?~

actually I feel like it makes sense to fix this now and worry about what we're doing with this sp after the spikes are complete

mehansen (approved):

ericnagel (approved):

PR #837, opened 2026-05-20
Fix sp_d_morbidity_report_postprocessing: rewrite user-comment join via NRT staging CSV

nullflux:
## Description
The user-comment query at lines 802-816 of `016-sp_nrt_morbidity_report_postprocessing-001.sql` joined the morb Order observation to itself and then filtered `obs_domain_cd_st_1 IN ('C_Order','C_Result')`. This is impossible, since the Order row's domain is `'Order'`. The temp table was always empty and `MORB_RPT_USER_COMMENT` never populated.

This PR rewrites it to walk Order to C_Result via the staging CSV `nrt_morbidity_observation.followup_observation_uid` (which the NRT layer already projects from the act_relationship graph), then filter `#morb_obs_reference` to `obs_domain_cd_st_1 = 'C_Result'` for the user-comment row's `activity_to_time`, `add_user_id`, and linked text. 

Verified locally by truncating `MORB_RPT_USER_COMMENT`, ran the SP for a morb Order from the ODSE generated test fixtures (uid=20080010), with table then showing the expected row. `job_flow_log` steps 19 and 27 at row_count=1, no errors.

## Related Issue

## Additional Notes
No `testData/unit` fixture — exercising the SP needs ~10 seeded tables across both DBs. End-to-end repro is in `utilities/comparison-fixtures/bugs/03_morb_rpt_user_comment/repro.sql`.

## Checklist


mpeels:
Thoughts on adding a unit test to cover this scenario?

mehansen:
can you explain what is meant by staging CSV?

ericbuckley:
> can you explain what is meant by staging CSV?

I think its just that the UID column is an intermediate comma separated list of values, not a single UID.

`string_split(rtrim(ltrim(nmo.followup_observation_uid)), '','') AS followupObs`

mehansen:
have you been able to repro this in the UI? I haven't been able to get data into this table via either MasterETL or RTR. the data dictionary says this table is for comments on an "externally entered Morbidity Report" so maybe there's some non-UI route for morb reports?
I see you referenced an end-to-end repro sql but I can't find utilities/comparison-fixtures/bugs/03_morb_rpt_user_comment/repro.sql

edit: 🕳️ 🐇 
Hailey showed me how to add a morb report user comment!
you need to create an external user and then have that external user create a morb report. morb reports created by external users have a comment functionality
Screenshot 2026-05-22 at 3 46 46 PM
however it's not working for me with RTR on this branch or main, only with MasterETL

nullflux:
> have you been able to repro this in the UI? I haven't been able to get data into this table via either MasterETL or RTR. the data dictionary says this table is for comments on an "externally entered Morbidity Report" so maybe there's some non-UI route for morb reports? I see you referenced an end-to-end repro sql but I can't find utilities/comparison-fixtures/bugs/03_morb_rpt_user_comment/repro.sql

@mehansen  I hadn't reproed this in the UI, only with generated / synthetic ODSE fixtures. That `repro.sql` doesn't exist on this branch. Thanks for including those UI steps from Haley, I'll perform those steps and compare the rows that it produces against what I've got in these fixtures, and then approach this bug from that vantage!

ericbuckley (approved):
TIL what CROSS APPLY does!

nullflux:
@mehansen Thanks again to both you and Hailey for documenting how to make this data in the UI!

There's one extra step that matters... after the external user submits, the report lands in the DRSA queue with no program area or jurisdiction. We have to open it as an internal user with `ASSIGNSECURITY` (e.g. `superuser`) and hit Transfer Ownership to assign a Program Area and Jurisdiction, which clears it out of DRSA. Thennn add the comment via the Add Comment button on the View page, that's the `MRB180` `C_Result` child.

As for why it worked in MasterETL but looked broken in RTR: the `MORB_RPT_USER_COMMENT` insert runs after the `MORBIDITY_REPORT_EVENT` insert in `sp_d_morbidity_report_postprocessing`, and that step needs a non-null `PATIENT_KEY` and an assigned report. A report still sitting in DRSA with no jurisdiction never completes that path, so the comment insert is never reached. Once the report is assigned out of DRSA, RTR on this branch _does_ populate the table. I confirmed an end-to-end UI capture producing the `MORB_RPT_USER_COMMENT` row in `RDB_MODERN`.

The updated fix should be right for the real data shape. I added a unit test (`unit/morbidityReportUserComment/`) based on the data captured from the trace, so it's covered going forward. (cc: @mpeels)

PR #842, opened 2026-05-21
Add sas container

mpeels:
## Description
Adds a SAS container to the docker compose as well as a local initialize script that:
1. Drops some tables to ensure a clean MasterETL run
2. Alter RDB log file size to allow for CDC capturing of MasterETL changes 
3. Enable CDC on RDB by default

## Related Issue
[APP-540](https://cdc-nbs.atlassian.net/browse/APP-540)

### Container info
The SAS container has `profiles: [sas]` so it will only start when called directly:
```sh
docker compose up sas -d
```

Or when the sas profile is included in the command
```sh
docker compose --profile sas up -d
```

To execute the MasterETL script the following 1 liner can be used
```sh
# Executes the MasterEtl script from outside the SAS container
docker compose exec -u SAS -it sas sh -c '/opt/wildfly-10.0.0.Final/nedssdomain/Nedss/BatchFiles/MasterEtl.sh'
```

To log into the SAS container and run it from within the following steps can be taken:

```sh
# Connect to SAS container as the SAS user
docker compose exec -u SAS -it sas bash

# Run MasterEtl script
/opt/wildfly-10.0.0.Final/nedssdomain/Nedss/BatchFiles/MasterEtl.sh
```


eliSkylight (approved):
Looks great!

containers/db/initialize/06-prep-for-masterEtl-trace.sql
```diff
@@ -0,0 +1,30 @@
...
+GO
+ALTER DATABASE RDB MODIFY FILE (NAME = 'RDB_log',  MAXSIZE = UNLIMITED, FILEGROWTH = 256MB);
+GO
+
+------------------------------------------------
+-- Enable CDC
+------------------------------------------------
+
+USE [RDB]
+EXEC sys.sp_cdc_enable_db;
```
ericbuckley, line 29: CDC on RDB to run SAS scripts?

mpeels: when tracing data to RDB, the first thing necessary is to enable CDC so the capture tool will do its thing. Adding the CDC enable to this init just eliminates the need to manually enable it each time we clear the DB.

It could be moved to its own init script or removed completely and we just have to manually enable it when needed. I'm fine with whatever

docker-compose.yaml
```diff
@@ -39,6 +39,25 @@ services:
       - "7003:7001"
       - "8788:8787"
 
+  sas:
+    image: ghcr.io/cdcent/nbs7-sas-linux:v1.0.4
+    profiles: [sas] # Only start if sas profile is enabled
```
hawley-skylight: make sure docs are up to date so folks know how to enable/disable sas profile?

eliSkylight (approved):

hawley-skylight (approved):

PR #849, opened 2026-05-26
Function Test IDs

ericnagel:
## Description
We're starting to have too many tests. Let's keep this file up-to-date with test IDs.

## Related Issue
N/A

## Additional Notes
N/A

## Checklist
- [X] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [X] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [X] I have updated the documentation, if applicable.
- [X] I have added or updated test cases to cover my changes, if applicable.

### DIFF
--- a/reporting-pipeline-service/src/test/resources/testData/functional/ids.md
+++ b/reporting-pipeline-service/src/test/resources/testData/functional/ids.md
@@ -1,0 +1,13 @@
+# Function Test IDs
+
+## IDs Used by Functional Tests
+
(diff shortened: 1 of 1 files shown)


hawley-skylight:
I agree that we should tidy the test suite up at this point, so that it is manageable as it is expanded. Why don't we chat about how to do this?

hawley-skylight (approved):

PR #852, opened 2026-05-27
Err/app 558/null patient race calc details d patient

eliSkylight:
## Description
Modified the `sp_patient_race_event` sql to reassign `PERSON_RACE_CALC_DETAILS` in a similar fashion to [PatientDimension.sas](https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/sas/dw/etl/src/PatientDimension.sas#L219) to account for `NULL` or empty values. Via UI, if the person race is selected as "Unknown" then the `D_PATIENT` table contains `NULL` for the column `PERSON_RACE_CALC_DETAILS` which is incorrect.

The existing bmirdCase functional test was updated to assign the race of "Unknown" to the target patient and validate the result on RDB_MODERN.

## Related Issue
[APP-558](https://cdc-nbs.atlassian.net/browse/APP-558)

## Additional Notes

## Checklist


ericnagel (approved):

liquibase-service/src/main/resources/db/005-rdb_modern/routines/053-sp_patient_race_event-001.sql
```diff
@@ -151,6 +151,15 @@ BEGIN
 --and cte1.rn = 1
         ;
 
+        --the sas PatientDimension.sas(line 219) for PATIENT_RACE_CALC_DETAILS column takes an extra step to reassign the column
+        --like so "IF LENGTHN(PATIENT_RACE_CALC_DETAILS)<1 THEN PATIENT_RACE_CALC_DETAILS='Unknown';". We should introduce
+         --a similar reassignment of the column to to prevent it from being NULL.
+
+        UPDATE sppr
```
ericbuckley, line 158: @eliSkylight why not just fix line 167?

eliSkylight: That line updates `PATIENT_RACE_CALCULATED` which does not have this discrepancy. The target for this PR is  `PATIENT_RACE_CALC_DETAILS`

ericbuckley: Ah I missed the difference between CALCULATED vs CALC_DETAILS.  However, if we update an empty value to 'Unknown' on line 159, won't that mean we miss the case of 'Unknown' on line 167?  In other words, while we are now correctly setting PATIENT_RACE_CALC_DETAILS when the value is missing, will that cause us to incorrectly set PATIENT_RACE_CALCULATED?

eliSkylight: I also confused the column names when I started looking at this sql so no worries! About missing the case of "Unknown" - my `expected.json` does make sure that both columns contain the value of `Unknown` when relevant. Regression testing also showed that `PATIENT_RACE_CALCULATED` was unaffected.

Additionally, as I understand it, the portion of the SQL that updates `PATIENT_RACE_CALCULATED` defaults to the value of `PATIENT_RACE_CALC_DETAILS `when there is only 1 race (in this case "Unknown"). Consider this condition in the stored procedure:

```sql
when CHARINDEX('|', PATIENT_RACE_CALC_DETAILS) = 0 then PATIENT_RACE_CALC_DETAILS
```
                  
  Here's the logical view of that condition when the value is "Unknown"
```shell
CHARINDEX('|','Unknown') = 0 => true => assigns PATIENT_RACE_CALCULATED = PATIENT_RACE_CALC_DETAILS => Unknown
```

ericbuckley: That sounds right, thanks for walking me through that and thanks for adding the test for both of those columns

reporting-pipeline-service/src/test/resources/testData/functional/bmirdCase/010-createPatient/expected.json
```diff
@@ -16,7 +16,9 @@
       "PATIENT_RECORD_STATUS": "ACTIVE",
       "PATIENT_STATE": "Georgia",
       "PATIENT_STATE_CODE": "13",
-      "PATIENT_UID": 12345544332211
+      "PATIENT_UID": 12345544332211,
+      "PATIENT_RACE_CALC_DETAILS": "Unknown",
+      "PATIENT_RACE_CALCULATED": "Unknown"
```
hawley-skylight: It was a good idea to add this to the existing functional test.

Let me know if you need help resolving the merge conflicts from the ID change.

hawley-skylight (approved):

reporting-pipeline-service/src/test/resources/testData/functional/bmirdCase/010-createPatient/expected.json
eliSkylight: Patrick,

I was just thinking about that. I think it would be safe for me to wait for your PR to merge then I'll pull it into my changes and adjust as needed!

liquibase-service/src/main/resources/db/005-rdb_modern/routines/053-sp_patient_race_event-001.sql
eliSkylight: My pleasure. I appreciate the closer look at the logic!

ericbuckley (approved):

PR #854, opened 2026-05-27
Functional Test of an ELR

ericnagel:
## Description
Primarily, this is a functional test of data coming into NBS via ELRs. Inbound data is held in a queue and then processed via the NBS User Interface. These steps are included in the various `setup.sql` files (ex: Transfer Ownership, Create Investigation, Associate with Investigation).

## Related Issue
[APP-570](https://cdc-nbs.atlassian.net/browse/APP-570)

## Additional Notes
`testing-tools/local-db-tracing/tracing_replay.py` has been updated to support the ELR User

`reporting-pipeline-service/src/test/resources/testData/functional/README.md` has been updated to reflect the next set of IDs, used by this test.

## Patient Story:
Taylor Swift experiences nausea and vomiting after drinking raw milk. She goes and sees her doctor who takes a stool sample.

1. Taylor's stool sample is tested and the following is received - 
    a. Condition = Shiga toxin-producing E. coli (STEC) 
    b. Test Type/Method = BioFire GI Panel 
    c. Result = detected (i.e. positive)
    d. Status = preliminary
2. Once the panel test is complete, the result will come in with the status marked as final. If I recall correctly, this ELR comes in as an update rather than a new lab report document in NBS. 
3. In addition to the panel that's done, the lab also performs a stool culture to confirm the pathogen by isolating STEC. There are 3 results that come in for this. Typically the way we'd see it is as follows - 
    a. 1 lab report created for the E. coli serotype - test type would be listed as stool culture and the result could be E. coli O157:H7 for example
    b. 1 lab report created for the shiga toxin results (there are 2) - shiga toxin 1 is detected (positive), shiga toxin 2 is not detected (negative) 
4. Because Taylor is a day care employee, she needs test negative in order to return to work. After her symptoms resolve, she has another specimen collected to do that additional testing. That ELR comes in as a PCR test and the result is indeterminate with the status as final. A few days later, a corrected ELR comes in with a result of negative and status = corrected.

## Checklist


reporting-pipeline-service/src/test/resources/testData/functional/elrEColi/010-createInvestigation/query.sql
```diff
@@ -0,0 +1,638 @@
+
+
+-- dbo.CASE_COUNT | operations: insert
+-- Query: 0
+-- Step: 1
+-- Logical comparison marked this identity as not comparison-safe.
```
mehansen: what do these comments mean?

ericnagel: I should remove those. 

When the query.sql is created, it doesn't really know how to query the data (the WHERE clause). I rewrote the where, but never removed these. I will - they mean nothing once they're in the functional test.

mehansen:
this is very cool
a non-blocking nice to have I think for future debugging/updating the test would be doc pointing out which actions or ELRs in the `functional-test-recordings` folder are used by which test steps

testing-tools/local-db-tracing/tracing_replay.py
mehansen: should the superuser id and the elr user ID be set in the same spot?

testing-tools/local-db-tracing/tracing_replay.py
```diff
@@ -25,6 +25,8 @@
 
 
 SUPERUSER_ID_VARIABLE = "@superuser_id"
+ELR_USER_ID_VARIABLE = "@elruser_id"
+DEFAULT_ELR_USER_ID = 10000015
```
mehansen: non-blocking question: any idea if this and the superuser ID always going to be these values or are these tools constrained to only work on the DB backup we've been using locally?

mehansen (approved):

testing-tools/local-db-tracing/tracing_replay.py
ericnagel: The tools are only meant to run against our DB backup, so the ID should not change.

testing-tools/local-db-tracing/tracing_replay.py
ericnagel: Good point; I moved both of these to the constants file.

PR #855, opened 2026-05-27
Enhancement: Functional Test Better Error Reporting

ericnagel:
## Description
When a functional test fails, it was sometimes difficult to understand why it was failing. This PR adds additional logging to help determine what went wrong. Example:

```
java.lang.AssertionError: Error executing query 0 in bmirdCase/010-createPatient/query.sql. SQL:
use RDB_MODERN
...
Caused by: org.springframework.jdbc.UncategorizedSQLException: PreparedStatementCallback; uncategorized SQLException for SQL [use RDB_MODERN]; SQL state [null]; error code [0]; The statement did not return a result set.
...
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: The statement did not return a result set.
...
Cause 1: org.springframework.jdbc.UncategorizedSQLException: PreparedStatementCallback; uncategorized SQLException for SQL [use RDB_MODERN]; SQL state [null]; error code [0]; The statement did not return a result set.
...
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: The statement did not return a result set.
```

## Related Issue
N/A

## Additional Notes
N/A

## Checklist

### DIFF
--- a/reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/support/QueryRunner.java
+++ b/reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/support/QueryRunner.java
@@ -50,17 +50,9 @@ public class QueryRunner {
     }
 
     return Stream.of(sql.split(";"))
-        .map(QueryRunner::stripLineComments)
         .map(String::trim)
+        .flatMap(statement -> statement.lines().map(String::trim).filter(line -> !line.isBlank() && !line.startsWith("--")))
         .filter(statement -> !statement.isBlank())
         .collect(Collectors.toList());
-  }
-
-  private static String stripLineComments(String statement) {
-    return statement
-        .lines()
-        .map(String::trim)
-        .filter(line -> !line.isBlank() && !line.startsWith("--"))
-        .collect(Collectors.joining(System.lineSeparator()));
   }
 }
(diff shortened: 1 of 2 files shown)


reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/support/QueryRunner.java
```diff
@@ -41,4 +43,24 @@ public static Optional<List<Map<String, Object>>> select(String query, JdbcClien
...
+        .collect(Collectors.toList());
+  }
+
+  private static String stripLineComments(String statement) {
+    return statement
+        .lines()
+        .map(String::trim)
+        .filter(line -> !line.isBlank() && !line.startsWith("--"))
+        .collect(Collectors.joining(System.lineSeparator()));
+  }
```
mpeels, line 62: Line 53 has a `String::trim` which is also present in the `stripLineComments` line 62. Similar to the filtering out of blank lines (55, and first of 63).

Can we do something like:

```java
public static List<String> splitStatements(String sql) {
    if (sql == null || sql.isBlank()) {
      return List.of();
    }

    return Stream.of(sql.split(";"))
        .map(String::trim)
        .filter(statement -> !statement.isBlank() && !line.startsWith("--"))
        .collect(Collectors.toList());
  }
```

and completely remove the helper method `stripLineComments`.


## Warning
Suggestion is untested and typed out while distracted

mpeels (approved):

reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/support/QueryRunner.java
ericnagel: Updated, thanks for the suggestion!

PR #858, opened 2026-05-28
APP-620: Dedupe nbs_case_answer in sp_s_pagebuilder_postprocessing

nullflux:
## Description
Single-select STD fields like `DM_INV_STD.CASE_DIAGNOSIS` could come out doubled
(`730 - Syphilis | 730 - Syphilis`). The `FOR XML PATH('')` concat in
`sp_s_pagebuilder_postprocessing` had no `DISTINCT`, so two `nbs_case_answer` rows with the
same answer text for the same question both got concatenated.

Added `DISTINCT` to both concat blocks (and switched the `ORDER BY` to the concatenated
expression, which `SELECT DISTINCT` requires). Real multi-selects are unaffected. Edited in
place since these routines are `runOnChange`.

Also added a unit test that seeds the duplicate case and checks `CASE_DIAGNOSIS` comes out
deduped — it fails on the old proc and passes on the fix.

## Related Issue
[APP-620](https://cdc-nbs.atlassian.net/browse/APP-620)

## Checklist


nullflux:
Reconsidering the `ORDER BY` change

nullflux:
Closing this out with no read-side change, per investigation of the originating NBS6 defect (NBS Central Redmine #17652, "Duplicated value in CASE_DIAGNOSIS … in DM_INV_STD", Component: MasterETL).

NBS6 resolved this via **source-data remediation** — `Duplicate_Answers_Check.sql` / `Duplicate_Answers_Delete.sql` that delete duplicate rows from `nbs_odse.dbo.NBS_case_answer` (dedup key `nbs_question_uid + answer_txt + act_uid [+ answer_group_seq_nbr]`, keeping `MIN(NBS_case_answer_uid)`). It was **not a stored-procedure change**, and those scripts were **not part of the 6.0.19 GA package**. NBS6's reporting ETL (`ETL_SP_STG_ALL`, which `sp_s_pagebuilder_postprocessing` is ported from) still carries the original un-`DISTINCT` `FOR XML` concat.

So the root cause is duplicate `NBS_case_answer` rows on the ingestion write path (not MERGE-ing), not the read-side concat. A read-side `DISTINCT`/dedup would diverge from NBS6 and risk reordering genuine multi-selects. The proper RTR fix, if pursued, belongs in the ingestion write path and should be tracked separately.

APP-620 closed as Done. Branch `aw/app-620` retained for reference.

PR #861, opened 2026-05-28
[APP-617] Adding LAB100.DOCUMENT_LINK

ericbuckley:
## Description
This pull request adds the `DOCUMENT_LINK` field to the `LAB_TEST` and `LAB100` tables, along with updates to the associated stored procedures and migration scripts. These enhancements enable the storage and retrieval of document URLs within the reporting database, improving accessibility to source documents.

## Related Issues
* [APP-617](https://cdc-nbs.atlassian.net/browse/APP-617)

## Additional Notes
* Created `LAB100` and `LAB_TEST` table structures via new Liquibase migration files.
* Updated `sp_d_lab_test_postprocessing-001.sql` and `sp_lab100_datamart_postprocessing-001.sql` to include the `DOCUMENT_LINK` field in insert and update operations.
* Added `DOCUMENT_LINK` to functional and unit test setup and expectation files to reflect the schema changes and verify data integrity.
* Added safe failure handling to migration scripts using `sqlcmd -b` and `set -e` (when one liquibase migration fails, the whole script should fail)

## Checklist


liquibase-service/migrate.sh
```diff
@@ -53,7 +55,7 @@ if [[ -n "$RUN_MIGRATIONS" && "$RUN_MIGRATIONS" == "true" ]]; then
     echo "Applying onboarding scripts"
     for sql in $(find "./02-onboarding" -iname "*.sql" | sort -V); do
         echo "Executing: $sql"
-        sqlcmd -C -S "${DB_HOST}" -U "${DB_USERNAME}" -P "${DB_PASSWORD}" -i "$sql"
+        sqlcmd -b -C -S "${DB_HOST}" -U "${DB_USERNAME}" -P "${DB_PASSWORD}" -i "$sql"
```
hawley-skylight, line 58: why add -b (and set -e)? do later onboarding scripts sometimes depend on earlier ones?

when running scripts 1 2 3 4 and 2 fails, I guess with this change it will not run 3 and 4, whereas
before it would run 3 and 4 even if 2 fails.

ericbuckley: `set -e` in the bash script will make the script fail immediately if any of the liquibase or sqlcmd calls result in a non-zero return code.

The `-b` flag to sqlcmd will terminate the batch job if any individual part fails.  I added this to protect against onboarding scripts that have multiple GO statements in them.

The goal was to not allow the "Migrations complete" statement to show up in the logs if any one of them fails.  We use that as a health check in our functional / unit tests.  If a migration fails, I think we should fail the tests.

hawley-skylight (approved):

PR #864, opened 2026-05-29
Fix stop/restart of mssql failing

mpeels:
## Description
Fixes an issue where stopping and starting the nbs-mssql container would fail due to tables already being deleted.


ericbuckley (approved):
❤️

PR #865, opened 2026-05-29
Remove unused dependency

mpeels:
## Description
After looking the debezium dependency PR created by dependabot: https://github.com/CDCgov/NEDSS-DataReporting/pull/859, I realized that the dependency is not necessary. The Java code does not interact directly with debezium.


ericbuckley (approved):
🪓

PR #866, opened 2026-05-29
APP-621 race condition

hawley-skylight:
## Description
F_INTERVIEW_CASE does not have a uniqueness constraint on D_INTERVIEW_KEY. Under certain race conditions,
that can result in duplicate D_INTERVIEW_KEYs in the F_INTERVIEW_CASE table. This pr avoids that happening
by updating sp_f_interview_case_postprocessing to hold a lock during the write to the F_INTERVIEW_CASE table. The nonclustered index is to avoid locking the entire table while the lock is held. 
Also included are three functional tests to test that the race condition is fixed. Verified locally that
the tests pass when the patch to sp_f_interview_case_postprocessing is applied, and fail when not applied.

## Related Issue
[APP-621](https://cdc-nbs.atlassian.net/browse/APP-621)

## Additional Notes

## Checklist


hawley-skylight:
Closed because this is not an issue fixed in 6.0.19.1 that needs fixing in NBS 7/RTR.

PR #868, opened 2026-06-01
Disable PHC datamart feature flag by default

mehansen:
## Description
Changes the default value of the `phc-datamart-enable` feature flag from `true` to `false` in the reporting pipeline service configuration to match what STLT will have in production. Aligning local and default config with production ensures we're validating prod behavior rather than a configuration that isn't live.

## Related Issue
[Slack convo](https://skylight-hq.slack.com/archives/C09KAHXEZT5/p1780339350420519)

## Additional Notes
The flag can still be enabled by setting `FF_PHC_DM_ENABLE=true`.

## Checklist


ericbuckley (approved):
wfm

PR #870, opened 2026-06-01
Fix bugs in patient and provider datamart sync stored procedures

mehansen:
## Description
Fixes multiple bugs in the stored procedures responsible for syncing patient/provider dimension columns to datamarts.
Note: Patient sync is functionally redundant but if we're leaving it in, which is the current direction, I thought it made sense to have working code and not have it logging errors. This was originally a small fix but as I was writing the unit test I kept discovering bugs that were preventing the SP from completing.

### Patient sync
* **VAR_DATAMART**: (1) `PATIENT_PHONE_NUMBER_HOME` was populated from `tmp.PATIENT_FIRST_NAME` instead of `tmp.PATIENT_PHONE_NUMBER_HOME`; (2) `PATIENT_AGE_REPORTED` and `PATIENT_WITHIN_CITY_LIMITS` mapped to nonexistent columns, causing the stored proc to fail with `Invalid column name` if there was an update for this datamart
* **TB_DATAMART** and **TB_HIV_DATAMART**: (1)`tb` alias on the right-hand side was referring to the table being updated, making every sync a no-op, and column names were wrong; (2) `p` alias was referring to a non-existent table; (3) CTE missing `FROM` clause
* **MORBIDITY_REPORT_DATAMART**: `RACE_CALCULATED` and `RACE_CALCULATED_DETAILS` mapped to nonexistent columns, causing the stored proc to fail with `Invalid column name 'RACE_CALCULATED'` if there was an update for this datamart

### Provider sync
* **AGGREGATE_REPORT_DATAMART**: `IF EXISTS` used `std_hiv_datamart_update = 1` but the update filtered on `tb_datamart_update = 1`; I believe `std_hiv_datamart_update` is the right one since [that's the one that checks if `PROVIDER_QUICK_CODE` was updated](https://github.com/CDCgov/NEDSS-DataReporting/blob/4824fe0b7005bcd1e5582ee21d1363b89a05c4d0/liquibase-service/src/main/resources/db/005-rdb_modern/routines/003-sp_nrt_provider_postprocessing-001.sql#L261-L266)

## Related Issue
found while investigating [APP-602](https://cdc-nbs.atlassian.net/browse/APP-602) and [APP-603](https://cdc-nbs.atlassian.net/browse/APP-603)

## Checklist


ericnagel:
Looks good; I think the related ticket is 603, not 602.

Is there a modification you could make to `reporting-pipeline-service\src\test\resources\testData\functional\morbidityReport\020-addMorbidityReport\setup.sql` so the test fails, then update `reporting-pipeline-service\src\test\resources\testData\functional\morbidityReport\020-addMorbidityReport\expected.json` (Case 3, `MORBIDITY_REPORT_DATAMART`) so it passes with your new logic?

mehansen:
@ericnagel 
> Looks good; I think the related ticket is 603, not 602.

I technically discovered it while I was reviewing the SQL for APP-602, but both investigation tickets did kind of blend together.

> Is there a modification you could make to `reporting-pipeline-service\src\test\resources\testData\functional\morbidityReport\020-addMorbidityReport\setup.sql` so the test fails, then update `reporting-pipeline-service\src\test\resources\testData\functional\morbidityReport\020-addMorbidityReport\expected.json` (Case 3, `MORBIDITY_REPORT_DATAMART`) so it passes with your new logic?

I think adding another step `030-updatePatient` would trigger the error but the wrinkle is that the morb report datamart is updated correctly anyway through a different pathway, so the only functional change is that we'll no longer log silent errors to the `job_flow_log`. That felt a little bit outside of the basic workflow(?) so I was planning on a unit test but I feel like I could go either way. What do you think?

ericnagel (approved):

ericbuckley (approved):

PR #881, opened 2026-06-09
Correctly populate INV_SUMM_DATAMART.NOTIFICATION_CREATE_DATE

ericnagel:
## Description

1. If there is no notification status → return NULL.
2. If there is a notification status:
    - First try RDB_DATE.DATE_MM_DD_YYYY.
    - If that is NULL, use the date portion of NOTIFICATION_LAST_CHANGE_TIME (time removed and cast back to DATETIME at midnight).

## Related Issue
[APP-607](https://cdc-nbs.atlassian.net/browse/APP-607)

## Additional Notes
Despite the alias being `NOTIFICATION_CREATE_DATE`, the fallback value is actually derived from `NOTIFICATION_LAST_CHANGE_TIME`, not a **create** timestamp. So if `RDB_DATE.DATE_MM_DD_YYYY` is missing, the column is really showing the notification's last-change date. This is not a change from the original code, but worth noting. There is not a better field to choose from in `[RDB_MODERN].[dbo].[NOTIFICATION]`

Once merged into main, revisit #817, validate the functional tests now pass, then merge 817 into main

## Checklist
- [x] I have added or updated test cases to cover my changes, if applicable. (PR-817 handles this)


ericbuckley (commented):
instead of NOTIFICATION_LAST_CHANGE_TIME, can we join on the nrt_investigation_notification table and use the `notif_add_time` value?

mehansen:
does this logic for how to populate the field match MasterETL or will there be diffs that we need to call out in the RTR reporting diffs doc?

ericnagel:
> does this logic for how to populate the field match MasterETL or will there be diffs that we need to call out in the RTR reporting diffs doc?

@mehansen that's what I'm _trying_ to figure out now. It's a **slow** process.

PR #887, opened 2026-06-10
Fix HIV Functional Test

ericnagel:
## Description
The main change here is updating a join on the HIV Functional Test, but I also added some code that allows me to specify an array of directories to only run those tests.

## Related Issue
N/A

## Checklist


reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/functional/DataDrivenFunctionalTests.java
```diff
@@ -30,6 +32,20 @@
...
+   * <p>Examples:
+   *
+   * <ul>
+   *   <li>List.of("hivNotificationActualReferral")
+   *   <li>List.of("interview", "elrEColi")
+   * </ul>
+   */
+  // private static final List<String> SELECTED_TEST_NAMES =
+  // List.of("hivNotificationActualReferral");
+  private static final List<String> SELECTED_TEST_NAMES = List.of();
```
ericbuckley, line 47: ❤️ this is great.   What do you think about adding an example to the README.md to show how to use this param

ericbuckley (approved):
Weird that it was working sometimes with that broken join clause.  Thanks for fixing!

reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/integration/functional/DataDrivenFunctionalTests.java
ericnagel: I'm going to check with @mpeels as I cannot locate documentation on the test suite, but the inline documentation works nicely:
image

ericbuckley: Maybe I misread, is this a parameter you can pass to gradlew to just run one or two functional tests and not all of them?  In which case, I was thinking we could add documentation / example here https://github.com/CDCgov/NEDSS-DataReporting/blob/main/README.md#testing

PR #894, opened 2026-06-11
APP-734: fix sld_investigation_repeat key alloc and TEXT pivot

nullflux:
## Description
sp_sld_investigation_repeat_postprocessing left D_INVESTIGATION_REPEAT empty. New rows got the sentinel key 1 and were filtered out, and the dynamic TEXT pivot NULL-propagated so TEXT columns stayed null. This fixes both, with a data-driven test that seeds a Pertussis investigation with repeating-block answers and asserts the rows land with populated TEXT.

## Related Issue
[APP-734](https://cdc-nbs.atlassian.net/browse/APP-734)

## Additional Notes
Covers APP-734 (key allocation) and the related TEXT-pivot defect in the same SP.

## Checklist


liquibase-service/src/main/resources/db/rdb/routines/010-sp_sld_investigation_repeat_postprocessing-001.sql
```diff
@@ -1141,7 +1148,30 @@ ON #CODED_TABLE_REPT
                     ON [PRIMARY];
             END;
 
+        -- APP-734 fix: re-seed the IDENTITY counter on D_REPT_KEY so the
```
ericbuckley:
```suggestion
        -- APP-734 fix: re-seed the IDENTITY counter on LOOKUP_TABLE_N_REPT so the
```

We're getting the counter from D_REPT_KEY, but we're reseeding LOOKUP_TABLE_N_REPT right?

ericbuckley (approved):

liquibase-service/src/main/resources/db/rdb/routines/010-sp_sld_investigation_repeat_postprocessing-001.sql
nullflux: Good catch, the comment was misleading. D_REPT_KEY is the IDENTITY column and LOOKUP_TABLE_N_REPT is the table it lives on, so the reseed is DBCC CHECKIDENT on LOOKUP_TABLE_N_REPT and the next auto-generated D_REPT_KEY is what flows into D_INVESTIGATION_REPEAT_KEY. The @reseed_to value is read from MAX(D_INVESTIGATION_REPEAT_KEY) so we land above the sentinel and any existing key. Reworded the comment to name both in 0778a28e.

PR #907, opened 2026-06-16
[APP-729] Refactor Kafka and Debezium connector initialization

ericbuckley:
## Description
This PR refactors the Kafka connector / Debezium registration process, migrating it from infrastructure-level shell scripts (`healthcheck.sh`) into the `reporting-pipeline-service` itself. The application now manages the automatic registration of required Debezium and Kafka Connect connectors upon startup. This approach centralizes configuration management, reduces complexity in Docker containers, and improves the reliability and maintainability of the connector deployment pipeline.

**NOTE:** Changes will need to be made to the reporting-pipeline-helm chart before deploying this code to a k8.  Specifically, we need to add ENV variables for DEBEZIUM_CONNECT_URL and KAFKA_CONNECT_URL.

## Related Issues
- [APP-729](https://cdc-nbs.atlassian.net/browse/APP-729)

## Additional Notes
This change replaces custom initialization scripts and volume mounts with a centralized, application-managed approach using Spring Boot components.

   - Introduced `ConnectorClient` to manage REST API interactions with Kafka Connect/Debezium.
   - Added `DebeziumConfig` and `KafkaConnectConfig` implementing `ApplicationRunner` to register connectors at startup.
   - Updated `docker-compose.yaml` to remove initialization logic and simplify service configurations.
   - Added unit and integration tests for the new connector management components.

## Checklist


reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorProperties.java
```diff
@@ -0,0 +1,22 @@
...
+
+  @Data
+  public static class Group {
+    private boolean enabled;
+    private String url;
+    private int retryAttempts = 20;
+    private long retryDelayMs = 5000;
+    private List<String> definitions = List.of();
+  }
+}
```
mpeels: Highly recommend using a `record` instead of a `@Data` annotated class unless there is some specific reason not to.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorProperties.java
```diff
@@ -0,0 +1,32 @@
+package gov.cdc.nbs.report.pipeline.connector;
+
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.ConstructorBinding;
+import org.springframework.boot.context.properties.bind.DefaultValue;
+
+@ConfigurationProperties(prefix = "connector")
```
mpeels, line 8: Info only: Instead of adding `@EnableConfigurationProperties(ConnectorProperties.class)` to the `ReportingPipelineServiceApplication`, I believe you can just put a `@Configuration` on this class. 

I can see benefits to doing it both ways and hold no strong preference.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorClient.java
```diff
@@ -0,0 +1,91 @@
...
+
+    String existing = restTemplate.getForObject(baseUrl + "/connectors", String.class);
+    if (existing != null && existing.contains("\"" + name + "\"")) {
+      log.info("Connector '{}' already registered at {}, skipping", name, baseUrl);
+      return;
+    }
+
+    restTemplate.put(baseUrl + "/connectors/" + name + "/config", config);
+    log.info("Registered connector '{}' at {}", name, baseUrl);
+  }
```
mpeels, line 86: Future work?: It would be nice to not only register if its missing but register if the connector's value was different.

I can definitely see a case where the app is started up with an improper config, then that bad config gets registered. Not an awful thing to have to delete the "bad" connector and then restart the app with the proper config, but would be nice to just fix the config and restart the app.

docker-compose.yaml
```diff
@@ -176,3 +164,6 @@ services:
         condition: service_healthy
       kafka-connect:
         condition: service_healthy
+
+volumes:
+  kafka-connect-plugins:
```
mpeels, line 169: Whats the benefit of having this  volume?

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorProperties.java
ericbuckley: We could decorate the ReportingPipelineServiceApplication with [ConfigurationPropertiesScan](https://docs.spring.io/spring-boot/api/java/org/springframework/boot/context/properties/ConfigurationPropertiesScan.html) to be less explicit about what we're configuring.  Thoughts?

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorClient.java
ericbuckley: Yeah, not to mention that a config could theoretically change from release to release (like we did when we enabled initial snapshot mode), so being able to re-register seems ideal.  I'll make a change.

docker-compose.yaml
ericbuckley: This vol is mounted to the kafka-connect service.  The thought was we can just use this to cache the jdbc driver we install.  It works, but honestly I'm not sure if it adds much, most of the startup time is spent in other processes, `confluent-hub install --no-prompt confluentinc/kafka-connect-jdbc:10.7.6` is quick.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorClient.java
ericbuckley: Looks like Kafka makes this simple, we can just remove the if-clause!  The PUT operation is a no-op if the config doesn't change.

https://github.com/CDCgov/NEDSS-DataReporting/pull/907/commits/d46d0dc871bc3eb1b93069794ded0b6ebf0bfed8

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/connector/ConnectorProperties.java
mpeels: Meh, I think I'm good as is.

mpeels (approved):

PR #926, opened 2026-06-25
[APP-765] nrt seeding health check

ericbuckley:
## Description
This pull request solves the seeding health check by introducing a new actuator endpoint to report on kafka lag; `/actuator/lag`.  This endpoint reports on `nbs_*` and `nrt_*` topics and tallies the approximate number of messages left.  With this information, users can obtain information as to when the initial seeding process is complete.

## Related Issues
[APP-765](https://cdc-nbs.atlassian.net/browse/APP-765)

## Additional Notes
We considered writing a set of queries to reconcile ODSE against RDB to confirm the nrt_* tables were seeded, but that would have been significantly more work and would have duplicated the filtering and transformation logic already encoded in the sp_*_event procedures — a brittle copy that would drift every time those procedures changed. Instead, we determine seeding by monitoring Kafka consumer-group lag: once the pipeline and sink groups have caught up to their topics, the seed is complete.

This PR also adds a small stored-procedure coverage module: it parses the migration SQL to build a catalog of every sp_* procedure and its steps, then scores that against the job_flow_log rows captured during the functional tests to report how many procedures (and steps) the suite exercised. It isn't required for this ticket, but it proved useful on a separate ticket for quantifying the stored-procedure coverage increase, and the report rides along in the existing test artifacts at no cost. It can be removed for clarity if preferred, but it's self-contained and harmless to leave in.

## Checklist


ericbuckley:
example screenshots of the `/actuator/lag` endpoint.

Screenshot 2026-06-26 at 16 00 08
Screenshot 2026-06-26 at 16 00 57

ericbuckley:
example stored proc coverage report (can also be obtained from functional test run in github actions).

[sproc-coverage.md](https://github.com/user-attachments/files/29398767/sproc-coverage.md)

ericnagel:
Where will this be documented? Anywhere in the repo that should be part of this PR?

mpeels (approved):

ericbuckley:
> Where will this be documented? Anywhere in the repo that should be part of this PR?

This particular [spring boot actuator](https://github.com/CDCgov/NEDSS-DataReporting/blob/main/reporting-pipeline-service/build.gradle#L99) endpoint will be documented in the SystemAdmin Guide, as we're going to ask users to monitor it while seeding is ongoing.

This of course [isn't the only one configured](https://github.com/CDCgov/NEDSS-DataReporting/blob/main/reporting-pipeline-service/src/main/resources/application.yaml#L132), so maybe adding a section to the README.md to document all of them is a good idea?  @ericnagel thoughts?

ericnagel:
> > Where will this be documented? Anywhere in the repo that should be part of this PR?
> 
> This particular [spring boot actuator](https://github.com/CDCgov/NEDSS-DataReporting/blob/main/reporting-pipeline-service/build.gradle#L99) endpoint will be documented in the SystemAdmin Guide, as we're going to ask users to monitor it while seeding is ongoing.
> 
> This of course [isn't the only one configured](https://github.com/CDCgov/NEDSS-DataReporting/blob/main/reporting-pipeline-service/src/main/resources/application.yaml#L132), so maybe adding a section to the README.md to document all of them is a good idea? @ericnagel thoughts?

If the others are documented in the SystemAdmin Guide, I'm good with this one being added there, too.

ericnagel (approved):

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/lag/KafkaAdminTopicLagReader.java
```diff
@@ -0,0 +1,96 @@
+package gov.cdc.nbs.report.pipeline.lag;
```
hawley-skylight, line 1: Not recommending this approach be changed as it does solve the problem. Just wondering if there could have been another simpler way to approach this: rather than reading offset information and computing, presumably Kafka can be configured (maybe it is already?) to provide aggregate metrics already.

reporting-pipeline-service/src/test/java/gov/cdc/nbs/report/pipeline/coverage/CoverageReport.java
```diff
@@ -0,0 +1,92 @@
+package gov.cdc.nbs.report.pipeline.coverage;
```
hawley-skylight, line 1: this is useful. thank you!

hawley-skylight (approved):

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/lag/KafkaAdminTopicLagReader.java
ericbuckley: There is the kafka metrics reporter, I'm not sure if that is enabled by default (so it could require an extra installation step on the STLTs side when they setup kafka), but to my knowledge that wouldn't be accessible from the apache kafka java SDK.

The only simpler route I could think of would be to use a third party library that provides similar functionality.  I'm always trying to balance "reinventing the wheel" and managing long-term dependency overhead. Given the high volume of CVEs popping up lately, adding a new library didn't seem justified for what amounts to [about 30 lines of code](https://github.com/CDCgov/NEDSS-DataReporting/pull/926/changes#diff-58435febd69adf57bee7d91fbe4dcc4c9231792516192211e9f03af2717bab10R50-R83).

However, if you have another idea on how to measure topic size, I'd be interested in hearing about it.

PR #930, opened 2026-06-26
Test updates after new Golden Backup

ericnagel:
## Description
Functional and Unit tests were failing after the new Golden Backup (due to duplicating data). This PR:

- Ensures conditions going into `nrt_condition_key` are unique / not duplicated
- Ensures conditions going into `CONDITION` are unique / not duplicated
- Adds a unique key to `condition_cd, program_area_cd` to prevent duplicates in the `nrt_condition_key` table
- Removes some fields from a functional test, that no longer exist
- Adds some empty values to a functional test - these were returned, but not expected so the test failed
- Adds some `DATE_KEY`s to a functional test
- Directs the tests to use the newer golden backup
- Merged in `/ph/APP-775-inv-summ-datamart-trigger` to fix the `INV_SUMM_DATAMART` failures
- Skipping tests of `L_INVESTIGATION_REPEAT` as the float as a UID is causing issues and this data should flow to downstream tests
- Explicitly name columns when populating nrt_srte_Condition_code (source has 33 columns while target has 35 columns in RDB_MODERN so `insert into dbo.nrt_srte_Condition_code select * from nbs_srte.dbo.Condition_code` was failing)

## Related Issue
[APP-730](https://cdc-nbs.atlassian.net/browse/APP-730)

## Additional Notes
Once merged in, PR #933 should be done soon after.

## Checklist


docker-compose.test.yaml
```diff
@@ -1,6 +1,6 @@
 services:
   nbs-mssql: !override
-    image: ghcr.io/cdcgov/nedss-datareporting-mssql:6.0.18.1
+    image: ghcr.io/cdcent/nedssdb:6.0.18.0-beta
```
mehansen: is this yaml getting auto applied in some way? I thought I saw that our functional and unit test runners were using the regular docker compose file

mehansen: oh ok iirc this file is for local development if you want to use the test db with the pre populated migrations via `CUSTOM_COMPOSE_FILE`

do we want the regular compose file to also use the new golden backup?

ericnagel: I think so, because things were failing, then I fixed them but I'll push both up to this branch and see what happens. Locally, I have both set to the beta.

mehansen: this is a good reminder that we should update this file with a new test DB image once the new baseline golden DB is working and finalized

ericnagel: @mehansen good thought; I opened PR #933 to update and publish this package, and also restore the `docker-compose.yaml` change.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/tables/252-create_nrt_condition_key-001.sql
```diff
@@ -23,3 +23,28 @@ IF NOT EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_condition_key' and xty
...
+
+IF EXISTS (SELECT 1 FROM sysobjects WHERE name = 'nrt_condition_key' and xtype = 'U')
+	AND NOT EXISTS (
+		SELECT 1
+		FROM sys.indexes
+		WHERE [name] = 'ux_nrt_condition_key_cd_prog_area'
+			AND [object_id] = OBJECT_ID('[dbo].nrt_condition_key')
+	)
+	BEGIN
+		CREATE UNIQUE INDEX ux_nrt_condition_key_cd_prog_area
```
ericbuckley, line 47: should we add an index to nrt_condition_key.CONDITION_KEY as well?

docker-compose.test.yaml
```diff
@@ -1,6 +1,6 @@
 services:
   nbs-mssql: !override
-    image: ghcr.io/cdcgov/nedss-datareporting-mssql:6.0.18.1
+    image: ghcr.io/cdcent/nedssdb:latest
```
ericbuckley: If we're going to revert to nedssdb:latest, should we just delete this file?  I'm not sure what value it provides over the main one.

ericbuckley (approved):

docker-compose.test.yaml
ericnagel: @ericbuckley I think there's a few files that can be deleted once we're happy with `6.0.17.0-replacement-beta`.
We should probably jump from `8a35ca` right to `6.0.17.0-replacement-beta` (tagged as latest)

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/tables/252-create_nrt_condition_key-001.sql
ericnagel: This is done above, [lines 35-36](https://github.com/CDCgov/NEDSS-DataReporting/pull/930/changes#diff-9486b0970695efd6753b4a78dac5cdafe8acba191766963bc16b9a9c6721983dR35)

hawley-skylight (approved):

eliSkylight (approved):

PR #935, opened 2026-06-29
APP-781: rank-align BMIRD strep pneumo additional-site datasets

nullflux:
## Description
When a Strep pneumo case has more than one value for one of the additional-site questions, the datamart was showing a single value repeated in every slot instead of the distinct sites that were entered. For example, non-sterile sites of "Amniotic fluid, Middle ear, Other" came out as "Amniotic fluid, Amniotic fluid, Amniotic fluid" (and likewise for the two additional-culture site columns).

The cause is in `sp_bmird_strep_pneumo_datamart_postprocessing`. It builds three separate site lists (non-sterile, and the two additional-culture lists) and then combined them by matching only on the investigation, which pairs every site against every other site (a Cartesian product). The pivot that follows then copied one value into all three slots.

The fix numbers each list per investigation (1st, 2nd, 3rd site) and lines the three lists up by that number, so the Nth site from each list ends up on the same row before being spread into the `_1/_2/_3` columns. A case with a single site is unaffected. Adds a data-driven regression test.

## Related Issue
[APP-781](https://cdc-nbs.atlassian.net/browse/APP-781)

## Additional Notes
This surfaced during the APP-471 synthetic-fixtures work, which is the first time the pipeline processed a case with multiple values for a site question, so the bug had gone unnoticed until now. Verified red-fix-green: the new test fails against the old routine and passes with the fix (`reporting-pipeline-service:test-unit -Dtests=app781_bmird_strep_pneumo_site`).

## Checklist

### DIFF
--- a/reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/routines/140-sp_bmird_strep_pneumo_datamart_postprocessing-001.sql
+++ b/reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/routines/140-sp_bmird_strep_pneumo_datamart_postprocessing-001.sql
@@ -631,113 +631,64 @@ BEGIN
 
         /*
         Site Data Processing -
-        Retrieve BMD125, BMD142 and BMD144 from BMIRD_MULTI_VALUE_FIELD and pivot into 3 columns
+        Retrieve BMD125, BMD142 and BMD144 from BMIRD_MULTI_VALUE_FIELD and pivot the top 3 entries into distinct columns.
 
-        Step 1: Create a dataset of all non-sterile sites
-        Step 2: Create 2 datasets of all additional culture sites
-        Step 3: Merge the datasets to create a new table with the columns transposed
-        Step 4: Merge the new table with the BMIRD_ANTIMICRO table
+        Step 1: Unpivot the non-sterile and additional culture sites from a single read into a unified row dataset using CROSS APPLY.
+        Step 2: Number the entries (1..N) per investigation and site type to establish their sequential order.
+        Step 3: Filter to the first 3 entries and pivot them directly into their final transposed columns (_1, _2, _3) using conditional aggregation.
+        Step 4: Merge the new table with the BMIRD_ANTIMICRO table.
+
+        Numbering each site type independently and pivoting by (site type, position) keeps the
+        Nth value of each question in its own slot. The previous version built three lists and
+        joined them on INVESTIGATION_KEY alone, a Cartesian product that, for any multi-answer
+        question, copied a single value into all three slots instead of the distinct sites.
         */
 
         BEGIN TRANSACTION;
 
         SET @Proc_Step_no = @Proc_Step_no + 1;
-        SET @Proc_Step_name = ' Generating #DM_BMD125, #DM_BMD142 and #DM_BMD144';
+        SET @Proc_Step_name = ' Generating #DM_BR7_T';
 
-        -- Build the three site lists for the case (non-sterile sites, and the two
-        -- additional-culture site lists). Number the entries 1..N per investigation
-        -- (rn) so that further down the lists can be lined up by position rather than
-        -- matching every value against every other value.
-        -- Step 1: non-sterile sites
-        SELECT INVESTIGATION_KEY, NON_STERILE_SITE_,
-               ROW_NUMBER() OVER (PARTITION BY INVESTIGATION_KEY ORDER BY NON_STERILE_SITE_) AS rn
-        into #DM_BMD125
-        FROM (
-            SELECT distinct bc.INVESTIGATION_KEY,
-                     a.NON_STERILE_SITE AS NON_STERILE_SITE_
+        WITH cte_combined AS (
+            SELECT DISTINCT
+                bc.INVESTIGATION_KEY,
+                v.Site_Type,
+                v.Site_Value
             FROM #BMIRD_PATIENT1 bc
-                 INNER JOIN dbo.BMIRD_MULTI_VALUE_FIELD a with (nolock)
-                            on bc.BMIRD_MULTI_VAL_GRP_KEY = a.BMIRD_MULTI_VAL_GRP_KEY
-            WHERE A.NON_STERILE_SITE IS NOT NULL
-        ) z;
-
-        -- Step 2: Create 2 datasets of all additional culture sites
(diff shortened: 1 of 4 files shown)


reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/routines/140-sp_bmird_strep_pneumo_datamart_postprocessing-001.sql
```diff
@@ -690,24 +700,30 @@ BEGIN
...
+                 UNION SELECT INVESTIGATION_KEY, rn FROM #DM_BMD144
+             ) k
+                 left outer join #DM_BMD125 a on a.INVESTIGATION_KEY = k.INVESTIGATION_KEY and a.rn = k.rn
+                 left outer join #DM_BMD142 b on b.INVESTIGATION_KEY = k.INVESTIGATION_KEY and b.rn = k.rn
+                 left outer join #DM_BMD144 c on c.INVESTIGATION_KEY = k.INVESTIGATION_KEY and c.rn = k.rn;
+
+        with cte2 as (
             SELECT *
-            FROM cte1 WHERE COUNTER <= 3
+            FROM #DM_BR7 WHERE COUNTER <= 3
```
ericbuckley: If we want to optimize here, we can skip the 4 temp previous temp tables and just insert directly into #DM_BR7.

```sql
        /*
        Site Data Processing -
        Retrieve BMD125, BMD142 and BMD144 from BMIRD_MULTI_VALUE_FIELD and pivot the top 3 entries into distinct columns.

        Step 1: Unpivot the non-sterile and additional culture sites from a single read into a unified row dataset using CROSS APPLY.
        Step 2: Number the entries (1..N) per investigation and site type to establish their sequential order.
        Step 3: Filter to the first 3 entries and pivot them directly into their final transposed columns (_1, _2, _3) using conditional aggregation.
        Step 4: Merge the new table with the BMIRD_ANTIMICRO table.
        */

        WITH cte_combined AS (
            SELECT DISTINCT
                bc.INVESTIGATION_KEY,
                v.Site_Type,
                v.Site_Value
            FROM #BMIRD_PATIENT1 bc
            INNER JOIN dbo.BMIRD_MULTI_VALUE_FIELD a WITH (NOLOCK)
                ON bc.BMIRD_MULTI_VAL_GRP_KEY = a.BMIRD_MULTI_VAL_GRP_KEY
            CROSS APPLY (VALUES 
                ('NON_STERILE_SITE', a.NON_STERILE_SITE),
                ('ADD_CULTURE_1_SITE', a.STREP_PNEUMO_1_CULTURE_SITES),
                ('ADD_CULTURE_2_SITE', a.STREP_PNEUMO_2_CULTURE_SITES)
            ) v (Site_Type, Site_Value)
            WHERE v.Site_Value IS NOT NULL
        ),
        cte_numbered AS (
            SELECT 
                INVESTIGATION_KEY,
                Site_Type,
                Site_Value,
                ROW_NUMBER() OVER (
                    PARTITION BY INVESTIGATION_KEY, Site_Type 
                    ORDER BY Site_Value
                ) AS COUNTER
            FROM cte_combined
        )
        SELECT INVESTIGATION_KEY,
               -- COUNTER = 1
               MAX(CASE WHEN Site_Type = 'NON_STERILE_SITE'   AND COUNTER = 1 THEN Site_Value END) AS NON_STERILE_SITE_1,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_1_SITE' AND COUNTER = 1 THEN Site_Value END) AS ADD_CULTURE_1_SITE_1,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_2_SITE' AND COUNTER = 1 THEN Site_Value END) AS ADD_CULTURE_2_SITE_1,
               -- COUNTER = 2
               MAX(CASE WHEN Site_Type = 'NON_STERILE_SITE'   AND COUNTER = 2 THEN Site_Value END) AS NON_STERILE_SITE_2,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_1_SITE' AND COUNTER = 2 THEN Site_Value END) AS ADD_CULTURE_1_SITE_2,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_2_SITE' AND COUNTER = 2 THEN Site_Value END) AS ADD_CULTURE_2_SITE_2,
               -- COUNTER = 3
               MAX(CASE WHEN Site_Type = 'NON_STERILE_SITE'   AND COUNTER = 3 THEN Site_Value END) AS NON_STERILE_SITE_3,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_1_SITE' AND COUNTER = 3 THEN Site_Value END) AS ADD_CULTURE_1_SITE_3,
               MAX(CASE WHEN Site_Type = 'ADD_CULTURE_2_SITE' AND COUNTER = 3 THEN Site_Value END) AS ADD_CULTURE_2_SITE_3
        INTO #DM_BR7_T
        FROM cte_numbered
        WHERE COUNTER <= 3
        GROUP BY INVESTIGATION_KEY;
```

nullflux: Dang, that's way better. Updated in e8d8ec3

reporting-pipeline-service/src/test/resources/testData/unit/app781_bmird_strep_pneumo_site/setup.sql
```diff
@@ -0,0 +1,68 @@
...
+    (22781011, 22781002, N'Middle ear',     NULL, NULL),
+    (22781012, 22781002, N'Other',          NULL, NULL),
+    (22781013, 22781002, NULL, N'Blood',                 NULL),
+    (22781014, 22781002, NULL, N'Bone',                  NULL),
+    (22781015, 22781002, NULL, N'Cerebral Spinal Fluid', NULL),
+    (22781016, 22781002, NULL, NULL, N'Joint Fluid'),
+    (22781017, 22781002, NULL, NULL, N'Lung'),
+    (22781018, 22781002, NULL, NULL, N'Sinus');
+
+EXEC dbo.sp_bmird_strep_pneumo_datamart_postprocessing @phc_uids = N'22781000', @debug = 0;
```
hawley-skylight, line 68: appreciate the unit test.

hawley-skylight (approved):

ericbuckley (approved):

PR #936, opened 2026-06-29
Add rdb-compare: local RDB vs RDB_MODERN comparison tool

nullflux:
## Description
Adds `rdb-compare`, a local tool for comparing the legacy **RDB** (MasterETL/SAS) against **RDB_MODERN** (the RTR pipeline). It matches rows on each table's business/UID key (never the surrogate `*_KEY` columns, which carry documented offsets), compares every other column NULL-aware via a cross-database join, and classifies each difference as **NEW** (unexplained, needs attention), **EXPECTED** (by design), **KNOWN_BUG**, or **IGNORED**. The known/expected differences live in a declarative rule registry (`rdb_compare/rules/`) rather than scattered if/else, so the report keeps the spotlight on the unexplained diffs. It emits a machine-readable `comparison.json` and a human-readable `comparison.md`.

It is the local-development counterpart to [CDCgov/NEDSS-DataCompare](https://github.com/CDCgov/NEDSS-DataCompare): the same key columns and comparison approach, but a single tool that reads two databases in one SQL Server instance, without the S3/Kafka/Spring machinery.

## Related Issue
[APP-720](https://cdc-nbs.atlassian.net/browse/APP-720)

## Additional Notes
This work was heavily influenced by @eliSkylight's comparison-testing work and his local testing tool.

Split out of the APP-471 synthetic ODSE fixtures work so it can be reviewed on its own. It is the comparison tool that pairs with that work, not the APP-471 PR itself. The engine, rule registry, classifier, and report renderers are unit-tested without a live database (`uv run pytest`); the DB layer is integration-tested against the dev SQL Server.

## Checklist

### DIFF
--- a/testing-tools/rdb-compare/README.md
+++ b/testing-tools/rdb-compare/README.md
@@ -1,0 +1,54 @@
+# rdb-compare
+
+A local tool that compares the legacy **RDB** (produced by MasterETL/SAS) against
+**RDB_MODERN** (produced by the RTR reporting pipeline), documenting which tables
+and which columns differ. It classifies every difference as **NEW**
+(unexplained, needs attention), **EXPECTED** (documented by-design),
+**KNOWN_BUG** (documented probable bug), or **IGNORED** (skip table, surrogate-key
+offset, ETL timestamp, audit column).
+
+It is the local-development counterpart to
+[CDCgov/NEDSS-DataCompare](https://github.com/CDCgov/NEDSS-DataCompare): it reuses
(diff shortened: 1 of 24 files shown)


ericbuckley (approved):
after this merges, what do you think about archiving the rdb compare repo and adding a link to indicate the work has been merged into DataReporting.

nullflux:
> after this merges, what do you think about archiving the rdb compare repo and adding a link to indicate the work has been merged into DataReporting.

I'm down! Curious what @eliSkylight thinks?

PR #943, opened 2026-07-02
COVID_CONTACT_DATAMART: Truncate and Repopulate

ericnagel:
## Description
The `COVID_CONTACT_DATAMART` was getting

```
n(n+1)
-------
  2
```
 
Records placed in it. Every time a contact was added, edited or deleted, ALL contacts were added to this table. This wasn't happening in MasterETL because the table was truncated then rebuilt. RTR would have rebuilt the table every time a COVID contact were added; this solution simply truncates the table, first.

This has been tested by:

1. Creating a patient
2. Creating a COVID investigation
3. Adding a contact (1 record in `COVID_CASE_DATAMART`)
4. Adding another contact (2 records)
5. Create another patient (B)
6. Create another COVID investigation (B)
7. Add a contact to investigation B (should be 3 records)
8. Editing the first contact (still 3 records)
9. Delete the second contact (back to 2 records)
10. Add another contact (3 records again)

## Related Issue
[APP-529](https://cdc-nbs.atlassian.net/browse/APP-529)

## Discussion
[Slack Thread](https://skylight-hq.slack.com/archives/C09KAHXEZT5/p1783003098393279)
[2nd Slack Thread](https://skylight-hq.slack.com/archives/C09KAHXEZT5/p1783018506794969)

## Additional Notes
This is a _TEMPORARY_ solution; a better solution would be to add indexes to the `COVID_CASE_DATAMART`

A functional test should also be created, which replicates the process described above.

## Checklist


reporting-pipeline-service/src/main/resources/db/changelog/migrations/v17.3/rdb/routines/315-sp_covid_contact_datamart_postprocessing-001.sql
```diff
@@ -399,11 +396,27 @@ BEGIN
         SET @proc_step_name = 'Update COVID_CONTACT_DATAMART';
         SET @proc_step_no = 2;
 
-        /* Start transaction for the delete and insert operations */
+        /* Start transaction for the truncate and insert operations */
         BEGIN TRANSACTION;
 
+        EXEC @LockResult = sys.sp_getapplock
```
hawley-skylight, line 402: what would be the negative consequence of not having a lock?

ericnagel: If lots of contacts are added, the table would be truncated and re-populated on top of itself. Imagine:

1. Contact F is added
2. COVID_CONTACT_DATAMART is truncated
3. Contact A, B, C are moved into the COVID_CONTACT_DATAMART
4. Now Contact G is added - while the previous sp is running
5. COVID_CONTACT_DATAMART is truncated
6. Contact D, E, F are added from previous sp run
7. Meanwhile Contacts A, B, C, D, E, F, G are added from new sp run
8. Therefore D, E, F are duplicated

With a lock, A-F would be populated but G would be missing (until Contact H is created)

This isn't the best long-term solution - it's the best solution today.

eliSkylight (approved):

hawley-skylight (approved):

ericnagel:
First Slack thread regarding this issue:

## Eric Nagel (BUF)  [10:38 AM]
Can someone please review https://cdc-nbs.atlassian.net/browse/APP-529 before I take this further?

The problem: `COVID_CONTACT_DATAMART` was just adding records, causing duplicates (and more)
There is no index on this table, so we can't (easily) ask, "Is this person already in the table?"
[The solution](https://github.com/CDCgov/NEDSS-DataReporting/compare/main...en/app-529-COVID_CONTACT_DATAMART) (without adding indexes, which I think is the right way to fix this) is to try to identify records by comparing values of various fields (I think I need to add more) and use the `NBS_ODSE.dbo.CT_contact_hist` and `NBS_ODSE.dbo.CT_contact` tables to identify the latest version of a user and their status (if deleted)
FWIW, the MasterETL solution is to wipe the whole table and rebuild it which works great for an overnight, batch process but does not work for RTR

I tested this by:

1. Creating a patient
2. Creating a COVID investigation
3. Adding a contact (1 record in COVID_CASE_DATAMART)
4. Adding another contact (2 records)
5. Editing the first contact (still 2 records)
6. Delete the second contact (back to 1 record)
7. Add another contact (2 records again)

Jira Cloud[Open](https://cdc-nbs.atlassian.net/browse/APP-529?atlOrigin=eyJpIjoiNDQ5N2UwYzVmZGM2NDJmNWI5ZTRkMTFmZTFiY2UxOGUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ&issueKey=APP-529&subProduct=jira-software)

## 26 replies

### Patrick Hawley (NYC)  [11:26 AM]
would it be possible to solve this with a separate table matching COVID_CASE_DATAMART_KEY to a contact key?
### Michael Peels  [11:29 AM]
ah, clever, using the _hist entry to find the row thats being updated.  It feels a bit hacky though.

If we did go with adding an index of some key to the table, what would that entail?
I guess we would need to do 1 full rebuild of the table on initial RTR setup to populate that new field and then adjust the stored proc that populates that table? (edited) 
### Eric Nagel (BUF)  [11:33 AM]
[@Patrick Hawley (NYC)](https://skylight-hq.slack.com/team/U09PMGJS4TY) the COVID_CASE_DATAMART has no key at all :disappointed:
[@Michael Peels](https://skylight-hq.slack.com/team/U09GQH1JQ8Y) definitely hacky...
the table may already have data so, yeah, adding a column would have to match the data to an ID (or, one time wipe and rebuild). Since I'm already identifying a contact based on data, I can easily populate their ID.
If I'm adding a contact key, I'd also want to add a PHC ID

### Patrick Hawley (NYC)  [12:36 PM]
? I see COVID_CASE_DATAMART_KEY in the schema
### Eric Nagel (BUF)  [12:39 PM]
sorry, COVID_CONTACT_DATAMART
### Patrick Hawley (NYC)  [1:35 PM]
Is there still any possibility of not changing this in RTR (feature not bug...) ? Fixing seems messy either on the value way, or other ways mentioned.
### Eric Nagel (BUF)  [1:37 PM]
If we do nothing, every time a contact is added, every contact will be re-added to the table.

Add a contact, table has 1 row
Add a 2nd contact, table now has 3 rows (1 + 2)
Add a 3rd contact, table now has 6 rows (3 + 3) (edited) 
[1:38 PM]With the code freeze on 7/6, this isn't going to make it into the upcoming release.
### Patrick Hawley (NYC)  [1:42 PM]
ugh
[1:43 PM]but if it does that, what will happen if the table is dropped when a contact is added? wildly inefficient, but would it be correct?
### Eric Nagel (BUF)  [1:43 PM]
I mean... my hacky fix can get into this release
[1:44 PM]The table isn't dropped - that behavior is only in MasterETL. But, that's another angle to take on this problem - if we agree it's temporary
[1:44 PM]Then it's 1 line to add, truncate then add everyone
### Michael Peels  [1:49 PM]
hacky fix for now, create ticket to do a better fix later?
### Eric Nagel (BUF)  [1:49 PM]
Which hacky fix? :laughing:
The one I did, or just drop a truncate in there and let all the records rebuild?
### Michael Peels  [1:50 PM]
the _hist one. I think there might be some issues with the truncate. especially at scale
[1:50 PM]but might be misunderstanding
### Eric Nagel (BUF)  [1:50 PM]
no I agree - at scale, truncate and rebuild will be bad
### Patrick Hawley (NYC)  [1:52 PM]
my inclination is to not do anything about this for this release. and to fix in a better way in next release-- like add key/index/column to the datamart as Eric initially suggested.
[1:55 PM]but I am concerned about behavior at scale if it is going to double the size of this table every time a contact is added. but I have a lot of questions of RTR behavior at scale already anyway.
### Eric Nagel (BUF)  [1:55 PM]
Just know that there will be

n ( n+1)
    2

 rows in that table.
[1:55 PM]Of all the options, I think that's the worst one.
### Patrick Hawley (NYC)  [1:56 PM]
another option: truncate the table once a day
### Eric Nagel (BUF)  [1:58 PM]
Daily won't help. If you add contacts 89 and 90, you'll have a table of 179 records. Which would then go to 0 overnight. Then up to 91 when that 91st contact is added
### Eric Nagel (BUF)  [2:06 PM]
I opened https://github.com/CDCgov/NEDSS-DataReporting/pull/940 - this is the "hacky match on data values" fix. I could get the "hacky truncate and rebuilt" fix done between meetings today. I really don't think we should ignore this bug
[2:06 PM]I have time after Retro to discuss

ericnagel:
Second Slack thread regarding this issue:

## Eric Nagel (BUF)  [2:55 PM]
I'd like to huddle here on APP-529 & PR-940 at 3:30pm ET (in 30ish min).

The options are:

1. Do nothing
2. Implement PR-940 (the hacky solution, matching on data fields)
3. Implement a new solution: truncate the table and insert all (the MasterETL solution)

(edited)
## 12 replies

### Eric Nagel (BUF)  [3:11 PM]
https://cdc-nbs.atlassian.net/browse/APP-529
https://github.com/CDCgov/NEDSS-DataReporting/pull/940
Jira CloudOpen

### A huddle happened  [3:30 PM]
You, Patrick Hawley (NYC), and 4 others were in the huddle for 28m.

### Eric Nagel (BUF)  [3:31 PM]
Starting a huddle on this topic, (in the thread)
[3:35 PM]@nbs-dragon anyone else want to listen in / give input?
### Michael Peels  [3:53 PM]
11:59
[3:57 PM]Does it get an entry in the job_batch_log by chance?
### Eric Nagel (BUF)  [3:57 PM]
yes
[3:58 PM]good idea
### Eric Buckley (PDX, he/him)  [4:01 PM]
SQL Server does have this function https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-ver17
learn.microsoft.comsys.sp_getapplock (Transact-SQL) - SQL Serversp_getapplock places a lock on an application resource.[4:02 PM]so you could attempt to get a lock, and if its not available, short-circuit

```sql
CREATE PROCEDURE dbo.my_protected_procedure
AS
BEGIN
    SET NOCOUNT ON;

    -- 1. Must be run inside a transaction if using @LockOwner = 'Transaction'
    BEGIN TRAN;

    DECLARE @LockResult INT;

    -- 2. Request an Exclusive lock on a custom string name
    EXEC @LockResult = sys.sp_getapplock
        @Resource = 'Mutex_my_protected_procedure_Lock',
        @LockMode = 'Exclusive',
        @LockOwner = 'Transaction',
        @LockTimeout = 0; -- 0 means: Exit immediately if already running.
                          -- Change to -1 if you want 2nd proc to wait in queue.

    -- 3. Evaluate if we successfully got the lock
    IF @LockResult < 0
    BEGIN
        -- Lock acquisition failed (Result -1 means it timed out / is already running)
        ROLLBACK TRAN;
        RAISERROR ('This stored procedure is already actively running in another session.', 16, 1);
        RETURN;
    END

    -- ===================================================
    -- YOUR CORE STORED PROCEDURE LOGIC GOES HERE
    -- ===================================================


    COMMIT TRAN;
END;
```

### Kasey Diebold (ATL)  [4:07 PM]
Decision: option 3; justification: aligns with requirements to match MasterETL, likely will have similar performance impact as existing behavior but duplicates will not be introduced; Notes: performance optimization will be needed in the future.
### Eric Nagel (BUF)  [4:49 PM]
I'll test this on Monday:
https://github.com/CDCgov/NEDSS-DataReporting/pull/943

(I added additional testing steps to cover multiple investigations)

PR #947, opened 2026-07-07
Improve NBS_Note query

mpeels:
## Description
Improves the `NBS_Note` query based on suggestions made in the Jira ticket.

## Related Issue
[APP-779](https://cdc-nbs.atlassian.net/browse/APP-779)

## Additional Notes
The unit test was run against the original stored procedure and the updated version. Both provide the same output.

There is a second part of the performance improvement that will involve a new migration to add an index to the `NBS_Note` table that will be a separate PR on the modernization repo. WIP - https://github.com/CDCgov/NEDSS-Modernization/pull/3395

## Checklist


reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/056-sp_investigation_event-001.sql
```diff
@@ -1011,17 +1011,20 @@ BEGIN
...
-                             GROUP BY note_parent_uid
+                 LEFT JOIN  (
+                              SELECT
+                                note_parent_uid,
+                                String_agg(CONVERT(VARCHAR(20), add_time) + '^' + note, '; ') AS PHC_NOTES
+                              FROM
+                                nbs_odse.dbo.nbs_note WITH (nolock)
+                              WHERE
+                                note_parent_uid IN (
+                                                      SELECT
```
ericbuckley, line 1022: This looks much better, I do wonder if we need to explicitly cast the split strings to an int to get the index to work, but maybe SQL Server is smart enough to make that optimization on its own?

eq
```sql
SELECT DISTINCT CONVERT(bigint, value) FROM STRING_SPLIT(@phc_id_list, ',')
```

ericbuckley (approved):
Looks good, just need to merge in main to resolve the v17.3 / v7.13 directory conflict.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/056-sp_investigation_event-001.sql
mpeels: better safe than sorry, will go ahead and add this

PR #948, opened 2026-07-07
Err/app 787/person service update nrt tables

eliSkylight:
**Description**

Routes `PersonService`'s patient/provider/auth_user postprocessing through a direct write to their `nrt_*` tables in RDB_MODERN, gated behind a `PERSON_SERVICE_DIRECT_WRITE` flag. Adds `NrtAuthUser`/`NrtPatient`/`NrtProvider` entities and repositories and functional tests covering the auth-user and provider direct-write paths (including a provider-update step). Architecture documentation has been created as well that lives in the person service module. Event messages are still published to the nrt_ per team recommendation.

**Related Issue**

[APP-787](https://cdc-nbs.atlassian.net/browse/APP-787)

**Additional Notes**

**Checklist**


reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/ARCHITECTURE.md
```diff
@@ -0,0 +1,102 @@
...
+    classDef muted fill:#16273d,stroke:#48607e,color:#6c839c,stroke-dasharray: 3 3;
+    class DW,JpaSave highlight;
+    class KafkaPublish,KafkaConnect,PPListener,ProcessNrt muted;
+```
+
+🟧 amber = `person-service-direct-write` (active path) · ⬜ dashed = Kafka-Connect (legacy, feature-flagged off)
+
+## Key conditionals
+
+- **`cd` field** routes `nbs_Person` events to patient vs. provider stored procs.
```
hawley-skylight: I thought the idea was to avoid the race between two things listening to the nrt_ topics (the sink connector, and the reporting-service) by ensuring that the nrt_ tables are always written first. 

In this diagram, it looks like the suggestion is to avoid the race by not writing to the nrt_ topics at all, in the Person case.

But how does this compare to this alternative:
 Before writing to the nrt_ topics, write to the nrt_ tables. Then send the message to the nrt_ topic, which is read by reporting-service and processed. (Turn off the sink connector for those topics, so there is no race as only one thing reads that topic.) The preserves some of the decoupling via messaging which may(?) help with scaling.

eliSkylight: If we publish to the nrt topics after writing to the nrt tables then Kafka-Connect will write to the nrt tables as well. We'll likely see errors in Kafka-Connect since the records would be there already (e.g., violation of KEY constraint) and who knows what issues that could cause.

Yes, we could modify the Kafka-Connect config to include the non-person nrt topics, but we would be forcing STLTs to perform additional manual processes so that they could switch between the 2 pathways.
1. Update the feature flag for direct write
2. Update the Kafka Connect config
3. Restart/Redeploy reporting-service-pipeline
4. Restart/Redeploy Kafka-Connect

hawley-skylight: Yes, the kafka connect configuration would need to be modified to stop writing.

The feature flag is not for users, but for RTR developers so that the code can be merged safely to the main branch while this work is under development. When the feature is fully completed and tested, the feature flag and old code paths can be removed.

eliSkylight: Makes sense to me. We discussed during the DRAGON demo day and there were different opinions on this. Perhaps a team review will help determine what is best.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/model/entity/NrtPatient.java
```diff
@@ -0,0 +1,280 @@
...
+
+  @Column(name = "race_white_gt3_ind")
+  private String raceWhiteGt3Ind;
+
+  @Column(name = "race_white_all")
+  private String raceWhiteAll;
+
+  private String patientNumber;
+  private String patientNumberAuth;
+  private String entryMethod;
```
mpeels, line 243: (nit, non-blocking): Some fields are annotated with `@Column` definitions and others are not. I personally prefer that all or none of them be annotated. 

I believe that the Spring Boot Hibernate implementation auto maps camel case to snake case (`patientNumber` becomes `patient_number`) by default. So we might not need any of the `@Column` annotations unless the field name does not match for some reason.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/ARCHITECTURE.md
mpeels: I might be speaking in ignorance here but I believe this statement is incorrect: 
```
We'll likely see errors in Kafka-Connect since the records would be there already 
(e.g., violation of KEY constraint)
``` 

I'm pretty sure the Kafka Sink Connector performs an upsert and will avoid any errors if data is already present.


I do like the idea of continuing to write to the `nrt_*` topics as the signal to drive post processing. This allows us to manage retry mechanisms and scaling at that level instead of having to restart processing by querying ODSE if something goes wrong.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/model/entity/NrtPatient.java
eliSkylight: makes sense to me. will update soon. Thanks!

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/ARCHITECTURE.md
eliSkylight: Makes sense @mpeels . It doesn't hurt to continue writing to the topics, but wondering if we need to determine if a config update is needed to handle duplicates because I believe a standard INSERT will cause the task to halt in Kafka-Connect if a key violation error is raised. https://docs.confluent.io/kafka-connectors/jdbc/current/sink-connector/overview.html#jdbc-sink-idempotent-writes

eliSkylight: @mpeels looks like we've accounted for that actually in **mssql-connector.json**: `"insert.mode": "upsert"`. Would you prefer then, that I remove the new enqueue method in PostProcessingService? TIA!

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/ARCHITECTURE.md
```diff
@@ -0,0 +1,106 @@
...
+
+    CdSwitch -->|PAT| ComputePatient["sp_Patient_Event"]
+    CdSwitch -->|PRV| ComputeProvider["sp_provider_event"]
+
+    subgraph DIRECTWRITE["Conditional: person-service-direct-write"]
+        DW{"directWrite flag"}
+        DW -->|true| JpaSave["JPA save<br/>→ nrt_patient / nrt_provider / nrt_auth_user"]
+        DW -->|false| KafkaPublish["Publish enriched JSON<br/>→ nrt.patient / nrt.provider / nrt.auth-user"]
+        JpaSave --> KafkaPublish
+        KafkaPublish -->|"directWrite=false only"| KafkaConnect["Kafka-Connect JDBC Sink<br/>upserts nrt_* table"]
```
mpeels: I believe with the latest changes, `false only` is no longer true.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/ARCHITECTURE.md
mpeels: Marking as resolved as I believe we have settled on going forward with the continued write to the `nrt_` topic.

mpeels (dismissed):

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/service/PersonService.java
```diff
@@ -229,6 +245,15 @@ private void processProviderData(List<ProviderSp> providerData) {
...
+            ProviderReporting reporting =
+                (ProviderReporting)
+                    transformer.processData(null, provider, PersonType.PROVIDER_REPORTING);
+            nrtProviderRepository.save(NrtProvider.from(reporting));
+            log.info(
+                "Provider data (uid={}) directly written to nrt_provider", provider.getPersonUid());
+          }
+
           String reportingKey = transformer.buildProviderKey(provider);
           String reportingData = transformer.processData(provider, PersonType.PROVIDER_REPORTING);
```
ericnagel: Leaving a note as we discussed moving this line up above the directWrite conditional and reusing it, instead of calling `transformer.processData` twice

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/service/PersonService.java
```diff
@@ -249,11 +246,9 @@ private void processProviderData(List<ProviderSp> providerData) {
               (ProviderReporting)
                   transformer.processData(null, provider, PersonType.PROVIDER_REPORTING);
 
-          if (directWrite) {
-            nrtProviderRepository.save(NrtProvider.from(providerReporting));
-            log.info(
-                "Provider data (uid={}) directly written to nrt_provider", provider.getPersonUid());
-          }
+          nrtProviderRepository.save(NrtProvider.from(providerReporting));
+          log.info(
```
ericnagel: Since direct write is the default, do we need this logging? (lines 295 and 340 as well)?

ericnagel (dismissed):

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/person/service/PersonService.java
eliSkylight: done! @ericnagel

PR #955, opened 2026-07-10
Set fixture last_chg_time dynamically at generation time

mehansen:
## Description
Synthetic ODSE fixture data previously hardcoded `last_chg_time` values in April 2026. Since the local dev DB backup image now includes a completed MasterETL run dated July 2, 2026, fixture rows were older than the last completed run and would not be picked up by a subsequent MasterETL execution.

This replaces the static `last_chg_time` literals across all fixture SQL files with `CAST(GETDATE() AS DATE)`, so that any newly generated fixture data is always timestamped as more recent than the last MasterETL run, regardless of when the dev DB image was last refreshed.

## Related Issue
[APP-808](https://cdc-nbs.atlassian.net/browse/APP-808)

## Additional Notes
- This fixes the issue dynamically going forward: every fixture load will stamp `last_chg_time` as "now" at generation/load time, so this should not need to be revisited the next time the dev DB backup image is refreshed with a newer MasterETL run.
- Verified: running MasterETL against freshly generated fixtures populates `RDB..LAB_TEST` with the same data as running on the old DB image.
- Used `CAST(GETDATE() AS DATE)` rather than `GETDATE()` directly so that every `last_chg_time` value resolves to midnight of the current day rather than the exact execution timestamp. Some fixtures have related rows spanning multiple tables where I wasn't sure whether `last_chg_time` should match across those related rows; truncating to midnight preserves that consistency regardless of how long the fixture load takes to run (unless you're running it right before midnight on the DB clock, which looks to be around 5pm PT). Open to switching back to `GETDATE()` if this causes issues or is unnecessarily confusing.

## Checklist


hawley-skylight (approved):
This is a problem with the functional tests too.

mpeels (approved):
Neato

PR #963, opened 2026-07-21
fix(debezium): support case-insensitive prefix drop

ericbuckley:
## Description
Update Debezium connector configurations to use case-insensitive and start-anchored regex for drop-prefix transforms, ensuring robust topic name routing.

## Checklist


hawley-skylight (approved):
why would some installations have differing capitalization for names?

ericbuckley:
> why would some installations have differing capitalization for names?

Most STLTs run a case-insensitive collation, so casing doesn't actually impact SQL Server queries — NBS_ODSE and nbs_odse behave the exact same way.

As to why it's different, probably because the NBS documentation / scripts haven't been consistent.  Even within RTR's documentation / scripts.  I've seen RDB_Modern and rdb_modern.

PR #966, opened 2026-07-28
set minimum max text repl size

ericbuckley:
## Description
Ensure SQL Server's instance-wide `max text repl size (B)` is at least
524,288 bytes (512 KiB) during bootstrap. This permits large LOB values
(such as `NBS_page.jsp_payload`) to enter CDC while keeping base64-encoded
Debezium messages within Kafka's default 1,048,588-byte limit. Existing
larger or unlimited values are preserved.

## Related Issue
[APP-883](https://cdc-nbs.atlassian.net/browse/APP-883)

## Additional Notes
Added a local/CI database initialization script that sets SQL Server’s max text repl size (B) to its 64 KiB default before the RTR bootstrap runs. This ensures containerized tests exercise and verify that the bootstrap raises the limit to 512 KiB. The initialization script preserves the existing show advanced options setting and does not affect production bootstrap behavior.

## Checklist

### DIFF
--- a/bootstrap/101-enable_cdc_on_odse_srte_databases-001.sql
+++ b/bootstrap/101-enable_cdc_on_odse_srte_databases-001.sql
@@ -7,7 +7,7 @@ IF IS_SRVROLEMEMBER('sysadmin') <> 1
 GO
 
 -- ------------------------------------------------------------
--- Configure the minimum LOB size that SQL Server can replicate
+-- Ensure SQL Server's replication LOB size limit is at least 512 KiB
 -- ------------------------------------------------------------
 -- This is an instance-wide setting. A 512 KiB binary value expands to about
 -- 683 KiB when Debezium's JSON converter base64-encodes it, leaving room for
@@ -21,8 +21,10 @@ SELECT @currentMaxTextReplicationSizeBytes = CONVERT(INT, VALUE_IN_USE)
 FROM SYS.CONFIGURATIONS
 WHERE NAME = 'max text repl size (B)';
 
-IF @currentMaxTextReplicationSizeBytes <> -1
-    AND @currentMaxTextReplicationSizeBytes < @minimumMaxTextReplicationSizeBytes
+IF
+    @currentMaxTextReplicationSizeBytes <> -1
+    AND @currentMaxTextReplicationSizeBytes
+    < @minimumMaxTextReplicationSizeBytes
     BEGIN
         SELECT @showAdvancedOptionsWasEnabled = CONVERT(BIT, VALUE_IN_USE)
         FROM SYS.CONFIGURATIONS
@@ -35,9 +37,9 @@ IF @currentMaxTextReplicationSizeBytes <> -1
             END
 
         PRINT 'Increasing max text repl size (B) from '
-            + CONVERT(VARCHAR(20), @currentMaxTextReplicationSizeBytes)
-            + ' to '
-            + CONVERT(VARCHAR(20), @minimumMaxTextReplicationSizeBytes);
+        + CONVERT(VARCHAR(20), @currentMaxTextReplicationSizeBytes)
+        + ' to '
+        + CONVERT(VARCHAR(20), @minimumMaxTextReplicationSizeBytes);
 
         EXEC SYS.SP_CONFIGURE
             'max text repl size (B)',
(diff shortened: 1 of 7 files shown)


containers/db/initialize/006-set-max-text-repl-size.sql
```diff
@@ -0,0 +1,38 @@
...
+USE [master];
+GO
+
+DECLARE @showAdvancedOptionsWasEnabled BIT;
+
+SELECT @showAdvancedOptionsWasEnabled = CONVERT(BIT, VALUE_IN_USE)
+FROM SYS.CONFIGURATIONS
+WHERE NAME = 'show advanced options';
+
+IF @showAdvancedOptionsWasEnabled = 0
```
ericnagel: Since you're testing if you can set 'show advanced options,' should this script also:

1. Get the initial value
2. If it's 1, set it to 0; if it's 0, leave it alone
3. Then set it to 1
4. Set the max text repl size
5. Set 'show advanced options' back to original value (if original wasn't 1, because that's where we're at now)

Basically, you're already making sure you can change 'max text repl size' but if 'show advanced options' starts at 1, you never check to make sure it, too, can be updated.

ericbuckley: I'm not sure I follow.

Are you saying if `show advanced options` is set to **1** before this script runs, we don't check to see if `max text repl size` can be changed?

ericnagel: As I understand it, `containers/db/initialize/006-set-max-text-repl-size.sql` sets `max text repl size` to a low value and checks to make sure that update was successful, so we know that `bootstrap/101-enable_cdc_on_odse_srte_databases-001.sql` will work.

But we never check to see if `show advanced options` can be changed. Maybe we don't need to?

mehansen:
this lgtm and seems to be working for me locally based on the container logs
have you been able to run the containerized tests locally? I wanted to validate that they exercise the path that would trigger the exception without the bootstrap update but I'm having issues and I think I need to fix my setup. happens on main as well so not related to your change.

containers/db/initialize/006-set-max-text-repl-size.sql
ericbuckley: @ericnagel I simplified containers/db/initialize/006-set-max-text-repl-size.sql.  Its just for our golden backup image, and we already have lots of assumptions in place for scripts 001-005.  If we know its always going to be 1, no need to check.

ericbuckley:
> this lgtm and seems to be working for me locally based on the container logs have you been able to run the containerized tests locally? I wanted to validate that they exercise the path that would trigger the exception without the bootstrap update but I'm having issues and I think I need to fix my setup. happens on main as well so not related to your change.

I have not, I also have problems running the functional and stored proc unit tests locally.

bootstrap/101-enable_cdc_on_odse_srte_databases-001.sql
```diff
@@ -6,6 +6,60 @@ IF IS_SRVROLEMEMBER('sysadmin') <> 1
     END
 GO
 
+-- ------------------------------------------------------------
+-- Configure the minimum LOB size that SQL Server can replicate
```
mehansen: nit but this should say max right?

edit: the minimum max? minimum limit? 😅

mehansen (approved):
this lgtm!

do our tests update `nbs_page.jsp_payload`? I haven't been able to confirm locally but that is nonblocking. either way I think having the config set to the default for our local DB is a good idea.

ericbuckley:
> this lgtm!
> 
> do our tests update `nbs_page.jsp_payload`? I haven't been able to confirm locally but that is nonblocking. either way I think having the config set to the default for our local DB is a good idea.

I don't think so, but thats a good idea for a future functional test

bootstrap/101-enable_cdc_on_odse_srte_databases-001.sql
ericbuckley: I was thinking min, because we need the value to be at least this large, or else CDC will fail.  If the users already have a repl size set, and its larger than the min of 512KiB, we don't touch anything

mehansen: this jumped out at me I think because the config we're changing is for the maximum LOB size that the SQL server can replicate, but we are setting a minimum for that config.

maybe this is a little more clear?
`-- Ensure SQL Server's replication LOB size limit is at least 512 KiB`

PR #967, opened 2026-07-29
Performance: sp_patient_event

mpeels:
## Description
Addresses some performance issues in the `sp_patient_event` and the `sp_patient_race_event` (called by `sp_patient_event`).

Adds the updated stored procedures as new migrations in the `v7.13` liquibase release

## Related Issue
[APP-911](https://cdc-nbs.atlassian.net/browse/APP-911)

### Rewrite of sp_patient_event details
Analysis of the execution plan from a production like environment showed that the `sp_patient_event`'s final SELECT statement was responsible for ~47% of the entire query's cost and that cost was concentrated in ~15 chained Nested Loop Left Joins - one per call to `fn_get_value_by_cd_ques`. 

1. Function Call Optimization
`fn_get_value_by_cd_ques` is called 12 times per person (11 on Person columns, 1 on Person_name.nm_suffix). Each call independently re-joins nbs_question -> codeset -> code_value_general (plus the DEM139/DEM142 UNION branches to naics_industry_code / language_code). That's replaced with a single UNPIVOT -> JOIN ONCE -> PIVOT pipeline (see the person_codes / person_code_desc / person_code_pivot CTEs) that reproduces the exact same lookup logic as the function, but as one set-based pass instead of 12 correlated calls. This removes the repeated Index Spools the optimizer was building for each call, and removes the serial-execution / non-parallelizable restriction that scalar/UDX function calls impose.

2. Temporary Table Filtering
The Person.code columns are filtered into a #filtered_person temp table up front (matching the original WHERE clause), and both the code-lookup pipeline and the final SELECT read from it, so the STRING_SPLIT + Person filter is only evaluated once, not once per CTE reference.

3. STRING_SPLIT Output Casting
STRING_SPLIT output is explicitly CAST to BIGINT to remove an implicit conversion flagged in the plan (CONVERT_IMPLICIT(bigint, STRING_SPLIT.value)).

4. Inlining Function Calls
`fn_get_record_status` is now INLINED as a plain CASE expression instead of a function call. Its definition is a simple deterministic mapping over a literal input with no table access, so there's no behavioral risk in inlining it, and it was the specific call the plan named as the reason the whole statement was forced to run single-threaded (NonParallelPlanReason = TSQLUserDefinedFunctionsNotParallelizable). With this and item 1 above both fixed, there is no longer any scalar/table UDF call left in the main patient SELECT that would block parallelism. dbo.fn_get_value_by_cvg and dbo.fn_get_user_name are now also INLINED as OUTER APPLYs in the FROM clause (preferred_gender_lookup, add_user_lookup, last_chg_user_lookup), replicating their source exactly. Both were called once per row rather than once per coded column, so their individual cost was smaller than item 1's pattern - but with these done too, there is no remaining scalar or table-valued function call anywhere in this proc's main patient SELECT.

### Rewrite of sp_patient_race_event details
Rewrote sp_patient_race_event to replace per-category, row-by-row (correlated-subquery) processing with a single set-based pass. Logic and output shape are unchanged; only the execution strategy changed.

1. Consolidated 5 category temp tables into 1
Previously, each of the 5 OMB race categories (AMER_IND, BLACK, WHITE, ASIAN, NAT_HI) got its own SELECT ... INTO temp table, each independently scanning #TMP_S_PERSON_RACE. These are now built in a single pass into #TMP_S_PERSON_RACE_CAT, tagged with a RACE_CATEGORY_TAG column via CASE.

2. Replaced 20 correlated-subquery UPDATEs with 1 windowed pivot
The original computed each category's _1/_2/_3/_4 slot columns via a separate CTE + UPDATE per slot per category (5 categories × 4 slots = 20 near-identical UPDATE statements), each of which re-joined back out to `NBS_ODSE.dbo.person_race` and `NBS_SRTE.dbo.Race_code`. The new version computes ROW_NUMBER() OVER (PARTITION BY patient, category) once across all categories together, then pivots ranks 1–4 into columns with a single MAX(CASE WHEN rn = n THEN ...) aggregation — no re-joins to source tables needed since the temp table already has everything.

3. Replaced STUFF/FOR XML PATH string concatenation with STRING_AGG
PATIENT_RACE_ALL, PATIENT_RACE_CALC_DETAILS, and each category's _ALL column were built using the classic STUFF((SELECT ... FOR XML PATH('')), 1, 1, '') pattern — an implicit per-row correlated subquery plus a separate temp table and SUBSTRING cleanup step to strip the leading delimiter. These are now built with STRING_AGG(...) WITHIN GROUP (ORDER BY ...), a native set-based aggregate, eliminating the correlated subqueries, the extra temp tables (#TEMP_*_RACE_ALL), and the substring post-processing entirely.

4. Collapsed 5 sequential UPDATE...JOINs into 1 pivoted UPDATE
Populating the final output row per patient previously required 5 separate UPDATE ... FROM ... JOIN statements, one per category temp table. This is now one UPDATE driven by a single cat_summary CTE that pivots all 5 categories' columns via conditional aggregation in one pass.

5. Removed the DISTINCT-driven duplicate handling
The original used SELECT DISTINCT to collapse duplicated root-race rows into one row per patient. The new version uses ROW_NUMBER() OVER (PARTITION BY patient_uid ...) = 1, which is generally cheaper than a DISTINCT over a wide row set and makes the one-row-per-patient intent explicit.

6. Added indexes on temp tables
Clustered indexes were added on PATIENT_UID (and PATIENT_UID + RACE_CATEGORY_TAG for the category table) on each intermediate temp table, supporting the joins/updates that follow rather than relying on unindexed scans.

Net Effect
Eliminates roughly 20 redundant re-joins back to `NBS_ODSE.dbo.person_race` / `NBS_SRTE.dbo.Race_code` that the original repeated for every category/slot combination.
Replaces row-at-a-time XML-based string building with native set-based aggregation.
Reduces the number of temp tables from ~11 down to 4, and the number of UPDATE statements from ~27 down to ~6.
No change to business logic, output columns, or error handling — this is a pure performance/maintainability refactor, so results should be identical for a given input batch.

## Checklist


mpeels:
Doing a little branch cleanup -- will return shortly

PR #970, opened 2026-07-29
Performance: sp_patient_event

mpeels:
## Description
Addresses some performance issues in the `sp_patient_event` and the `sp_patient_race_event` (called by `sp_patient_event`).

Adds the updated stored procedures as new migrations in the `v7.13` liquibase release

## Test note
The patient event test was updated to include a lot more race data prior to deploying the stored procedure changes.  The test was verified as passing both with and without the updated stored procedures.

## Related Issue
[APP-911](https://cdc-nbs.atlassian.net/browse/APP-911)

### Rewrite of sp_patient_event details
Analysis of the execution plan from a production like environment showed that the `sp_patient_event`'s final SELECT statement was responsible for ~47% of the entire query's cost and that cost was concentrated in ~15 chained Nested Loop Left Joins - one per call to `fn_get_value_by_cd_ques`. 

1. Function Call Optimization
`fn_get_value_by_cd_ques` is called 12 times per person (11 on Person columns, 1 on Person_name.nm_suffix). Each call independently re-joins nbs_question -> codeset -> code_value_general (plus the DEM139/DEM142 UNION branches to naics_industry_code / language_code). That's replaced with a single UNPIVOT -> JOIN ONCE -> PIVOT pipeline (see the person_codes / person_code_desc / person_code_pivot CTEs) that reproduces the exact same lookup logic as the function, but as one set-based pass instead of 12 correlated calls. This removes the repeated Index Spools the optimizer was building for each call, and removes the serial-execution / non-parallelizable restriction that scalar/UDX function calls impose.

2. Temporary Table Filtering
The Person.code columns are filtered into a #filtered_person temp table up front (matching the original WHERE clause), and both the code-lookup pipeline and the final SELECT read from it, so the STRING_SPLIT + Person filter is only evaluated once, not once per CTE reference.

3. STRING_SPLIT Output Casting
STRING_SPLIT output is explicitly CAST to BIGINT to remove an implicit conversion flagged in the plan (CONVERT_IMPLICIT(bigint, STRING_SPLIT.value)).

4. Inlining Function Calls
`fn_get_record_status` is now INLINED as a plain CASE expression instead of a function call. Its definition is a simple deterministic mapping over a literal input with no table access, so there's no behavioral risk in inlining it, and it was the specific call the plan named as the reason the whole statement was forced to run single-threaded (NonParallelPlanReason = TSQLUserDefinedFunctionsNotParallelizable). With this and item 1 above both fixed, there is no longer any scalar/table UDF call left in the main patient SELECT that would block parallelism. dbo.fn_get_value_by_cvg and dbo.fn_get_user_name are now also INLINED as OUTER APPLYs in the FROM clause (preferred_gender_lookup, add_user_lookup, last_chg_user_lookup), replicating their source exactly. Both were called once per row rather than once per coded column, so their individual cost was smaller than item 1's pattern - but with these done too, there is no remaining scalar or table-valued function call anywhere in this proc's main patient SELECT.

### Rewrite of sp_patient_race_event details
Rewrote sp_patient_race_event to replace per-category, row-by-row (correlated-subquery) processing with a single set-based pass. Logic and output shape are unchanged; only the execution strategy changed.

1. Consolidated 5 category temp tables into 1
Previously, each of the 5 OMB race categories (AMER_IND, BLACK, WHITE, ASIAN, NAT_HI) got its own SELECT ... INTO temp table, each independently scanning #TMP_S_PERSON_RACE. These are now built in a single pass into #TMP_S_PERSON_RACE_CAT, tagged with a RACE_CATEGORY_TAG column via CASE.

2. Replaced 20 correlated-subquery UPDATEs with 1 windowed pivot
The original computed each category's _1/_2/_3/_4 slot columns via a separate CTE + UPDATE per slot per category (5 categories × 4 slots = 20 near-identical UPDATE statements), each of which re-joined back out to `NBS_ODSE.dbo.person_race` and `NBS_SRTE.dbo.Race_code`. The new version computes ROW_NUMBER() OVER (PARTITION BY patient, category) once across all categories together, then pivots ranks 1–4 into columns with a single MAX(CASE WHEN rn = n THEN ...) aggregation — no re-joins to source tables needed since the temp table already has everything.

3. Replaced STUFF/FOR XML PATH string concatenation with STRING_AGG
PATIENT_RACE_ALL, PATIENT_RACE_CALC_DETAILS, and each category's _ALL column were built using the classic STUFF((SELECT ... FOR XML PATH('')), 1, 1, '') pattern — an implicit per-row correlated subquery plus a separate temp table and SUBSTRING cleanup step to strip the leading delimiter. These are now built with STRING_AGG(...) WITHIN GROUP (ORDER BY ...), a native set-based aggregate, eliminating the correlated subqueries, the extra temp tables (#TEMP_*_RACE_ALL), and the substring post-processing entirely.

4. Collapsed 5 sequential UPDATE...JOINs into 1 pivoted UPDATE
Populating the final output row per patient previously required 5 separate UPDATE ... FROM ... JOIN statements, one per category temp table. This is now one UPDATE driven by a single cat_summary CTE that pivots all 5 categories' columns via conditional aggregation in one pass.

5. Removed the DISTINCT-driven duplicate handling
The original used SELECT DISTINCT to collapse duplicated root-race rows into one row per patient. The new version uses ROW_NUMBER() OVER (PARTITION BY patient_uid ...) = 1, which is generally cheaper than a DISTINCT over a wide row set and makes the one-row-per-patient intent explicit.

6. Added indexes on temp tables
Clustered indexes were added on PATIENT_UID (and PATIENT_UID + RACE_CATEGORY_TAG for the category table) on each intermediate temp table, supporting the joins/updates that follow rather than relying on unindexed scans.

Net Effect
Eliminates roughly 20 redundant re-joins back to `NBS_ODSE.dbo.person_race` / `NBS_SRTE.dbo.Race_code` that the original repeated for every category/slot combination.
Replaces row-at-a-time XML-based string building with native set-based aggregation.
Reduces the number of temp tables from ~11 down to 4, and the number of UPDATE statements from ~27 down to ~6.
No change to business logic, output columns, or error handling — this is a pure performance/maintainability refactor, so results should be identical for a given input batch.

## Checklist


reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/rdb.changelog-7.13.yaml
```diff
@@ -3287,3 +3287,21 @@ databaseChangeLog:
         - sqlFile:
             path: db/changelog/migrations/v7.13/rdb/tables/262-alter_nrt_dead_letter_log.sql
             splitStatements: false
+  - changeSet:
+      id: app-911-sp_patient_event-performance
+      author: liquibase
+      runOnChange: true
+      changes:
+        - sqlFile:
+            path: db/changelog/migrations/v7.13/rdb/routines/371-sp_patient_event.sql
```
mpeels: ah, so it would be `371-sp_patient_event-002.sql`?  -- Edit: `054-sp_patient_event-002.sql`

mpeels: no strong feelings either way. I'm not sure how valuable the `revision` numbering is.

mpeels: Just caught up on the slack thread. I do agree with the naming convention. Will update

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/053-sp_patient_race_event-001.sql
```diff
@@ -170,709 +169,211 @@ BEGIN
...
+
+        -- Flag whether the patient has more than 3 distinct sub-races
+        -- reported for this category (i.e. a 4th slot was populated).
+        update #TMP_S_PERSON_RACE_CAT
+        set RACE_GT3_IND =
                 case
-                    when [PATIENT_RACE_NAT_HI_4] is not null then 'TRUE'
-                    when [PATIENT_RACE_NAT_HI_4] is null then 'FALSE'
+                    when RACE_4 is not null then 'TRUE'
+                    when RACE_4 is null then 'FALSE'
```
ericnagel:
```suggestion
                    else 'FALSE'
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -98,17 +145,133 @@ BEGIN
          patient_uid_race_out)
             exec dbo.sp_patient_race_event @user_id_list, @batch_id;
```
ericnagel, line 156:
```suggestion
        -- ---------------------------------------------------------------
        -- Materialize the incoming person_uid list into a temp table so
        -- the subsequent Person lookup can use a relational join instead
        -- of a string-splitting subquery. This is generally more efficient
        -- for larger Person tables and keeps the filtering plan simpler.
        -- ---------------------------------------------------------------
        CREATE TABLE #requested_person_ids
        (
            person_uid BIGINT NOT NULL PRIMARY KEY
        );

        INSERT INTO #requested_person_ids (person_uid)
        SELECT CAST(value AS BIGINT) AS person_uid
        FROM STRING_SPLIT(@user_id_list, ',')
        WHERE TRY_CAST(value AS BIGINT) IS NOT NULL;
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -98,17 +145,133 @@ BEGIN
...
+        -- Pull the base Person records for only the requested person_uid
+        -- values, restricted to cd = 'PAT' (patient-type Person entities).
+        -- Materialized into a temp table (with a unique clustered index on
+        -- person_uid) so subsequent joins/CTEs perform well and person_uid
+        -- lookups are indexed.
+        -- ---------------------------------------------------------------
+        SELECT p.*
+        INTO #filtered_person
+        FROM nbs_odse.dbo.Person p WITH (NOLOCK)
+        WHERE p.person_uid IN (SELECT CAST(value AS BIGINT) FROM STRING_SPLIT(@user_id_list, ','))
```
ericnagel:
```suggestion
                 JOIN #requested_person_ids rpi ON rpi.person_uid = p.person_uid
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -159,62 +313,85 @@ BEGIN
...
                nested.email                                                                AS 'patient_email',
                nested.race                                                                 AS 'patient_race',
                nested.entity_id                                                            AS 'patient_entity'
-        FROM nbs_odse.dbo.Person p WITH (NOLOCK)
+        FROM #filtered_person p
+                 LEFT JOIN person_code_pivot pcp ON pcp.person_uid = p.person_uid
+
+                 -- Resolve preferred_gender_cd against the NBS_STD_GENDER_PARPT
+                 -- code set. TOP 1 + ORDER BY guards against duplicate code
+                 -- entries returning multiple rows.
```
ericnagel:
```suggestion
                 -- code set using the shared helper function. TOP 1 + ORDER BY
                 -- guards against duplicate code entries returning multiple rows.
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -159,62 +313,85 @@ BEGIN
...
+                 OUTER APPLY (
+                     SELECT TOP 1 cvg.code_short_desc_txt
+                     FROM nbs_srte.dbo.code_value_general cvg WITH (NOLOCK)
+                     WHERE cvg.code_set_nm = 'NBS_STD_GENDER_PARPT'
+                       AND cvg.code = NULLIF(p.preferred_gender_cd, '')
+                     ORDER BY cvg.code_short_desc_txt
```
ericnagel, line 368:
```suggestion
                     SELECT TOP 1 f.code_short_desc_txt
                     FROM dbo.fn_get_value_by_cvg(NULLIF(p.preferred_gender_cd, ''), 'NBS_STD_GENDER_PARPT') AS f
                     ORDER BY f.code_short_desc_txt
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -159,62 +313,85 @@ BEGIN
...
+
+                 -- Resolve the user who created the record ("Last, First").
+                 -- Guards against add_user_id <= 0 (system/no user) before joining.
+                 OUTER APPLY (
+                     SELECT TOP 1 CAST((RTRIM(LTRIM(au.user_last_nm)) + ', ' +
+                                        RTRIM(LTRIM(au.user_first_nm))) AS VARCHAR(150)) AS user_full_name
+                     FROM NBS_ODSE.dbo.Auth_user au WITH (NOLOCK)
+                     WHERE p.add_user_id > 0
+                       AND au.NEDSS_ENTRY_ID = p.add_user_id
+                     ORDER BY au.last_chg_time DESC
```
ericnagel, line 381:
```suggestion
                     SELECT u.user_full_name
                     FROM dbo.fn_get_user_name(p.add_user_id) AS u
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -159,62 +313,85 @@ BEGIN
...
+                 ) AS add_user_lookup(add_user_name)
+
+                 -- Resolve the user who last modified the record ("Last, First").
+                 OUTER APPLY (
+                     SELECT TOP 1 CAST((RTRIM(LTRIM(au.user_last_nm)) + ', ' +
+                                        RTRIM(LTRIM(au.user_first_nm))) AS VARCHAR(150)) AS user_full_name
+                     FROM NBS_ODSE.dbo.Auth_user au WITH (NOLOCK)
+                     WHERE p.last_chg_user_id > 0
+                       AND au.NEDSS_ENTRY_ID = p.last_chg_user_id
+                     ORDER BY au.last_chg_time DESC
```
ericnagel, line 393:
```suggestion
                     SELECT u.user_full_name
                     FROM dbo.fn_get_user_name(p.last_chg_user_id) AS u
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/054-sp_patient_event-001.sql
```diff
@@ -290,17 +483,31 @@ BEGIN
...
+                                                    OUTER APPLY (
+                                                        SELECT TOP 1 cvg.code_short_desc_txt
+                                                        FROM nbs_odse.dbo.nbs_question nq WITH (NOLOCK)
+                                                                 JOIN nbs_srte.dbo.codeset cs WITH (NOLOCK)
+                                                                      ON cs.code_set_group_id = nq.code_set_group_id
+                                                                 JOIN nbs_srte.dbo.code_value_general cvg WITH (NOLOCK)
+                                                                      ON cvg.code_set_nm = cs.code_set_nm
+                                                        WHERE nq.question_identifier = 'DEM107'
+                                                          AND cvg.code = NULLIF(pn.nm_suffix, '')
+                                                        ORDER BY cvg.code_short_desc_txt
```
ericnagel, line 520:
```suggestion
                                                        SELECT TOP 1 f.code_short_desc_txt
                                                        FROM dbo.fn_get_value_by_cd_ques(NULLIF(pn.nm_suffix, ''), 'DEM107') AS f
                                                        ORDER BY f.code_short_desc_txt
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/053-sp_patient_race_event-001.sql
```diff
@@ -24,143 +66,100 @@ BEGIN
...
+                   WHEN '2028-9' THEN 'ASIAN'     -- Asian
+                   WHEN '2076-8' THEN 'NAT_HI'    -- Native Hawaiian / Other Pacific Islander
+                   END AS RACE_CATEGORY_TAG
+        into #TMP_S_PERSON_RACE_CAT
         FROM #TMP_S_PERSON_RACE
-        WHERE RACE_CATEGORY_CD = '2028-9'
+        WHERE RACE_CATEGORY_CD IN ('1002-5', '2054-5', '2106-3', '2028-9', '2076-8')
           AND RACE_CD <> RACE_CATEGORY_CD
---ORDER BY PATIENT_UID
         ;
```
ericnagel:
```suggestion
        ;WITH race_category_map AS (
            SELECT '1002-5' AS race_category_cd, 'AMER_IND' AS race_category_tag
            UNION ALL SELECT '2054-5', 'BLACK'
            UNION ALL SELECT '2106-3', 'WHITE'
            UNION ALL SELECT '2028-9', 'ASIAN'
            UNION ALL SELECT '2076-8', 'NAT_HI'
        )
        SELECT sr.*, 
               m.race_category_tag AS RACE_CATEGORY_TAG
        INTO #TMP_S_PERSON_RACE_CAT
        FROM #TMP_S_PERSON_RACE sr
                 JOIN race_category_map m
                      ON sr.RACE_CATEGORY_CD = m.race_category_cd
        WHERE sr.RACE_CD <> sr.RACE_CATEGORY_CD;
```

This would be easier to maintain - no need to add the `race_category_cd` in 2 places if this is expanded upon later.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/053-sp_patient_race_event-001.sql
```diff
@@ -24,143 +66,101 @@ BEGIN
         where pr.person_uid in (SELECT value FROM STRING_SPLIT(@user_id_list, ','))
```
ericnagel: I missed a `@user_id_list` , sorry
 
`where pr.person_uid in (SELECT value FROM STRING_SPLIT(@user_id_list, ','))` will take every `pr.person_uid` (`[bigint]`) and convert it to a string in order to compare against the split `@user_id_list`

PR #979, opened 2026-08-05
[APP-904] reduce job_flow_log volume with configurable event-procedure logging

ericbuckley:
## Description
Made routine job_flow_log writes optional across RTR event stored procedures, defaulting logging off to reduce database and transaction-log volume. Errors remain logged unconditionally, while existing diagnostics, result sets, transactions, nested calls, and positional compatibility are preserved. The EVENT_PROCEDURE_DEBUG_LOGGING setting is propagated through all relevant Java call paths and can be enabled temporarily or for coverage runs. SQL-backed tests, coverage checks, and operational documentation were updated accordingly.

## Related Issues
- [APP-904](https://cdc-nbs.atlassian.net/browse/APP-904)

## Additional Notes
- sp_patient_race_event is intentionally unchanged because it only writes unconditional errors.
- The new setting controls application-invoked routine logging; direct SQL callers must pass @debug_logging = 1 explicitly when needed.
- Functional coverage enables logging automatically, as to capture accurate coverage stats.

## Checklist

### DIFF
--- a/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/repository/ContactRepository.java
+++ b/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/repository/ContactRepository.java
@@ -8,6 +8,11 @@ import org.springframework.data.repository.query.Param;
 
 public interface ContactRepository extends JpaRepository<Contact, String> {
 
-  @Query(nativeQuery = true, value = "exec sp_contact_record_event :ct_contact_uid")
-  Optional<Contact> computeContact(@Param("ct_contact_uid") String contactUid);
+  @Query(
+      nativeQuery = true,
+      value =
+          "execute sp_contact_record_event @cc_uids = :ct_contact_uid, @debug = 0, "
+              + "@debug_logging = :debugLogging")
+  Optional<Contact> computeContact(
+      @Param("ct_contact_uid") String contactUid, @Param("debugLogging") boolean debugLogging);
 }
--- a/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
+++ b/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
@@ -448,7 +448,8 @@ public class ProcessInvestigationDataUtil {
       logger.info(ex.getMessage(), "InvestigationConfirmationMethod");
     } catch (Exception e) {
       logger.error(
-          "Error processing investigation confirmation method JSON array from investigation data: {}",
+          "Error processing investigation confirmation method JSON array from investigation data:"
+              + " {}",
           e.getMessage());
     }
   }
@@ -526,16 +527,12 @@ public class ProcessInvestigationDataUtil {
   }
 
   @Transactional(isolation = Isolation.REPEATABLE_READ)
-  public void processPhcFactDatamart(String publicHealthCaseUid) {
-    processPhcFactDatamart(publicHealthCaseUid, false);
-  }
-
-  @Transactional(isolation = Isolation.REPEATABLE_READ)
   public void processPhcFactDatamart(String publicHealthCaseUid, boolean debugLogging) {
     try {
       // Calling sp_public_health_case_fact_datamart_event
       logger.info(
-          "Executing stored proc: sp_public_health_case_fact_datamart_event '{}' to populate PHС fact datamart",
+          "Executing stored proc: sp_public_health_case_fact_datamart_event '{}' to populate PHС"
+              + " fact datamart",
           publicHealthCaseUid);
       investigationRepository.populatePhcFact(publicHealthCaseUid, debugLogging);
       logger.info(
@@ -551,7 +548,8 @@ public class ProcessInvestigationDataUtil {
     try {
       // Calling sp_public_health_case_fact_datamart_update
       logger.info(
-          "Executing stored proc: sp_public_health_case_fact_datamart_update '{}', '{}' to update PHС fact datamart",
+          "Executing stored proc: sp_public_health_case_fact_datamart_update '{}', '{}' to update"
+              + " PHС fact datamart",
           objName,
           uid);
       investigationRepository.updatePhcFact(objName, uid);
@@ -609,7 +607,8 @@ public class ProcessInvestigationDataUtil {
 
     } catch (Exception e) {
       logger.error(
-          "Error processing Investigation Interview or any of the associated data from interview data: {}",
+          "Error processing Investigation Interview or any of the associated data from interview"
+              + " data: {}",
           e.getMessage());
     }
   }
--- a/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
+++ b/reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
@@ -527,12 +527,17 @@ public class ProcessInvestigationDataUtil {
 
   @Transactional(isolation = Isolation.REPEATABLE_READ)
   public void processPhcFactDatamart(String publicHealthCaseUid) {
+    processPhcFactDatamart(publicHealthCaseUid, false);
+  }
+
+  @Transactional(isolation = Isolation.REPEATABLE_READ)
+  public void processPhcFactDatamart(String publicHealthCaseUid, boolean debugLogging) {
     try {
       // Calling sp_public_health_case_fact_datamart_event
       logger.info(
           "Executing stored proc: sp_public_health_case_fact_datamart_event '{}' to populate PHС fact datamart",
           publicHealthCaseUid);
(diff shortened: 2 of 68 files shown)


reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
```diff
@@ -527,12 +527,17 @@ private void processInvestigationAggregate(
 
   @Transactional(isolation = Isolation.REPEATABLE_READ)
   public void processPhcFactDatamart(String publicHealthCaseUid) {
+    processPhcFactDatamart(publicHealthCaseUid, false);
```
ericnagel: If this is always false, why not keep the existing code and send false along on line 540?

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/repository/ContactRepository.java
```diff
@@ -8,6 +8,11 @@
 
 public interface ContactRepository extends JpaRepository<Contact, String> {
 
-  @Query(nativeQuery = true, value = "exec sp_contact_record_event :ct_contact_uid")
-  Optional<Contact> computeContact(@Param("ct_contact_uid") String contactUid);
+  @Query(
+      nativeQuery = true,
+      value =
+          "execute sp_contact_record_event @cc_uids = :ct_contact_uid, @debug = 0, "
```
ericnagel, line 14: Is `@debug = 0` necessary here? Can you rely on the sproc's default parameter value?

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/051-sp_organization_event-001.sql
```diff
@@ -6,35 +6,38 @@ IF EXISTS (SELECT * FROM sysobjects WHERE  id = object_id(N'[dbo].[sp_organizati
...
-                 @batch_id
-               ,'Organization PRE-Processing Event'
-               ,'sp_organization_event'
-               ,'START'
-               ,0
-               ,LEFT('Pre ID-' + @org_id_list,199)
-               ,0
-               ,LEFT(@org_id_list,199)
-               );
+        IF @debug_logging = 1
```
ericnagel, line 19: Consider a new stored procedure whose only role is to write to the job_flow_log. Then you could call that proc every time and let it exit early if logging is disabled. This would reduce the lines of code in each procedure by 10 for every write to job_flow_log, decrease indentation, and centralize the logic

Example (doesn't align 1:1 here)
```sql
EXEC dbo.sp_log_job_flow
    @debug_logging = @debug_logging,
    @batch_id = @batch_id,
    @dataflow_name = @DATAFLOW_NAME,
    @package_name = @PACKAGE_NAME,
    @status_type = 'START',
    @step_number = 0,
    @step_name = LEFT('Pre ID-' + @treatment_uids, 199),
    @row_count = 0,
    @message_description = LEFT(@treatment_uids, 199);
```

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/repository/ContactRepository.java
ericbuckley: I originally thought so, but after more research it doens't appear to be required.  So its only being specified now to be explicit. Do you think its better to just let the default value be used?

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/util/ProcessInvestigationDataUtil.java
ericbuckley: I think we should just delete this, now that I'm looking deeper it was just added to satsify a test case.  nothing in the service calls this method (InvestigationService does call the one on line 534).  Deleting the test case and this method makes the most sense.

reporting-pipeline-service/src/main/java/gov/cdc/nbs/report/pipeline/investigation/repository/ContactRepository.java
ericnagel: Since it's already there, I don't think it's worth removing. But this could be a coding standard that we establish internally.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/051-sp_organization_event-001.sql
ericbuckley: @ericnagel I added [03f1907](https://github.com/CDCgov/NEDSS-DataReporting/pull/979/commits/03f1907f91ff5fef892b18c063798b8d2fc4e7de) and [7dc3fa4](https://github.com/CDCgov/NEDSS-DataReporting/pull/979/commits/7dc3fa431a2445f90988e6cf974391f6996d72b1) to test this out.  I can roll them back if you think its too much.  Either way, let me know.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/074-sp_add_job_flow_log-001.sql
```diff
@@ -0,0 +1,51 @@
...
+      AND OBJECTPROPERTY(id, N'IsProcedure') = 1
+)
+BEGIN
+    DROP PROCEDURE [dbo].[sp_add_job_flow_log]
+END
+GO
+
+CREATE PROCEDURE dbo.sp_add_job_flow_log
+    @batch_id BIGINT,
+    @dataflow_name VARCHAR(200),
```
ericnagel: This is nvarchar(199) in the current table

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/074-sp_add_job_flow_log-001.sql
```diff
@@ -0,0 +1,51 @@
...
+    @package_name VARCHAR(200),
+    @status_type VARCHAR(20),
+    @step_number FLOAT,
+    @step_name VARCHAR(200),
+    @row_count INT,
+    @msg_description1 VARCHAR(200) = NULL,
```
ericnagel: Also, [Msg_Description2] [nvarchar](500)

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/051-sp_organization_event-001.sql
ericnagel: @ericbuckley I'd also suggest sending @debug_logging in as an argument (defaults to 1, to maintain current functionality), and put your "if" (and early return) in the new stored procedure. This would remove the `IF @debug_logging = 1`, `BEGIN` and `END` from each proc call (-3 lines, +1 line (the new argument)), and outdent the `EXEC` line(s)

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/074-sp_add_job_flow_log-001.sql
ericbuckley: good catch

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/074-sp_add_job_flow_log-001.sql
ericbuckley: purposfully, I wanted the other params to be varchars as we don't need 2 bytes for each char with dataflow_name, package_name, etc.  To be safe, maybe we need to allow for unicode characters with msg_description1 and error_description, so I'll update those

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/051-sp_organization_event-001.sql
ericbuckley: I have mixed feelings on this one.  "debug_logging" is something I would typically leave off by default, then enable when I need more details.  Which is the case we are handling now with the sp_*_event procs.  Looking down the road though, if we try to use this method in PostProcessing stored procs, I think we should leave all the job_flow_log calls on all the time, with no option to turn the off (so we have an audit trail maintained).  I see how that will DRY the code up by a line or 2 in every sp_*_event proc, but I wonder if its worth it, since down the road someone could accidently turn off job_flow_logs for PostProcessing.

ericnagel: Ok, this isn't something I'm _that_ passionate about. I think the code is clearer your way, even if it has a couple extra lines of code (and 1 more indentation, my arch nemesis). I'm good with your method here.

ericnagel (approved):

PR #981, opened 2026-08-05
Performance: sp_lab100_datamart_postprocessing

mpeels:
## Description
Addresses performance issues of the `sp_lab100_datamart_postprocessing` stored procedure.

## Related Issue
[APP-922](https://cdc-nbs.atlassian.net/browse/APP-922)

## Bugs uncovered but not addressed
[APP-946](https://cdc-nbs.atlassian.net/browse/APP-946)
[APP-947](https://cdc-nbs.atlassian.net/browse/APP-947)

### 1. Unscoped full-table maintenance moved to a scheduled job
The old procedure's final two steps ran on **every batch invocation** regardless of batch size:
- Marking `LAB100` rows `INACTIVE` based on inactive orders
- An `EXCEPT`-based delete of `LAB100` rows no longer present in `LAB_TEST`

Both scanned the entirety of `LAB100`/`LAB_TEST`. These have been extracted into a separate scheduled job, so per-batch cost now scales with `@labtestuids` instead of full table size.

### 2. Reduced transaction scope
The previous version wrapped **every step** — including steps that only touch temp tables — in its own `BEGIN TRANSACTION` / `COMMIT TRANSACTION`. This procedure now runs all temp-table staging outside of any transaction and opens a single transaction scoped only around the two operations that need atomicity: the `UPDATE` and `INSERT` into `dbo.LAB100`.

- Removes dozens of unnecessary commits/log flushes per batch run.

### 3. Correlated subquery → set-based aggregate
`INVESTIGATION_KEYS` was previously built with a correlated subquery using `STUFF(... FOR XML PATH(''))` per group (effectively row-by-row). Replaced with:

```sql
STRING_AGG(CAST(investigation_key AS VARCHAR(20)), ', ')
    WITHIN GROUP (ORDER BY investigation_key)
```

### 4. Fewer temp table materializations
Several intermediate temp tables that existed only to stage a join or drop a duplicate column have been collapsed:

- `TMP_LABTEST_RESULTS_VAL` → `ALTER TABLE...DROP COLUMN` → `TMP_LABTEST_RESULTS_VAL_COMMENT` → `ALTER TABLE...DROP COLUMN` → `TMP_LABTEST_UPDATED`
  → collapsed into a single `#TMP_LABTEST_RESULT_ENRICHED`, built with an explicit column list instead of `SELECT *`. Avoids two full temp-table rewrites (`ALTER TABLE DROP COLUMN` is not metadata-only once data has been written) and removes a materialization step entirely.
- `TMP_LAB_REPORTING_ORG` / `TMP_ORDERING_ORG` (single-column projections used only to join to `D_ORGANIZATION`) removed — `TMP_LAB_ENTITY1` / `TMP_LAB_ENTITY2` now join directly from `#TMP_PERSON_ORDER_PROVIDER`.
- `TMP_LABTESTS` + `TMP_LABTESTS2` (two-pass LOINC resolution, the second pass re-projecting every column) merged into a single step using `CROSS APPLY`.

### 5. Cleaner parameter typing
`@debug` changed from `BIT = 'false'` (implicit string→bit conversion, compared against `'true'`/`'false'`) to `BIT = 0` (compared directly against `1`/`0`).

### 6. Dead code removed
Commented-out `NBS_ODSE.user_profile` join and associated unused `LAB_REPORT_LAST_UPDATED_BY` columns removed.

### 7. Supporting index on `dbo.LAB100.RESULTED_LAB_TEST_KEY`
Every DML statement in the procedure (`UPDATE`, `INSERT`'s anti-join, and now the scheduled job's `INACTIVE` update / orphan delete) matches rows against `RESULTED_LAB_TEST_KEY`, but `LAB100` had no supporting index for that predicate. Added:

```sql
CREATE NONCLUSTERED INDEX IX_LAB100_RESULTED_LAB_TEST_KEY
    ON dbo.LAB100 (RESULTED_LAB_TEST_KEY)
    INCLUDE (
        LAB_RPT_LOCAL_ID,
        RESULTED_LAB_TEST_CD_DESC,
        RESULTEDTEST_VAL_CD_DESC,
        NUMERIC_RESULT_WITHUNITS,
        LAB_RESULT_TXT_VAL,
        LAB_RESULT_COMMENTS,
        PATIENT_KEY,
        ELR_IND,
        SPECIMEN_COLLECTION_DT,
        LAB_RPT_RECEIVED_BY_PH_DT
    );
```

The `INCLUDE` columns cover the fields most commonly read alongside the key lookup, letting the update/match operations satisfy from the index without a key/RID lookup back to the base table. Creation is guarded by an existence check in `sys.indexes`, so the script is safe to re-run.

- Turns the previous-likely table/clustered-index scan on every `UPDATE`/`INSERT`/anti-join match into a seek.
- Directly benefits the now-scheduled `INACTIVE` update and orphan delete from item 1, since both also filter/join on `RESULTED_LAB_TEST_KEY`.

## Checklist


reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/019-sp_lab100_datamart_postprocessing-001.sql
```diff
@@ -1,828 +1,659 @@
...
+            ORDERING_PROVIDER_KEY
+        INTO #TMP_LABTEST_LABTESTRESULT
+        FROM dbo.LAB_TEST lt WITH (NOLOCK)
+            LEFT OUTER JOIN dbo.LAB_TEST_RESULT ltr WITH (NOLOCK)
+                ON lt.LAB_TEST_KEY = ltr.LAB_TEST_KEY
+        WHERE lt.LAB_TEST_KEY <> 1
+          AND lt.lab_test_uid IN (
+                SELECT lab_test_uid AS uid
+                FROM dbo.LAB_TEST WITH (NOLOCK)
+                WHERE lab_test_uid IN (SELECT value FROM STRING_SPLIT(@labtestuids, ','))
```
ericnagel, line 90: Consider using the temp table technique here, where uids are only converted once and an inner join on the temp table is used. This may or may not be faster, depending on many things.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/rdb.changelog-7.13.yaml
```diff
@@ -3287,3 +3287,20 @@ databaseChangeLog:
         - sqlFile:
             path: db/changelog/migrations/v7.13/rdb/tables/262-alter_nrt_dead_letter_log.sql
             splitStatements: false
+  - changeSet:
+      id: app-922-lab100_datamart_performance
+      author: liquibase
+      runOnChange: true
```
ericbuckley, line 3302:
```suggestion
      runOnChange: false
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/371-sp_lab100_cleanup.sql
```diff
@@ -0,0 +1,192 @@
...
+                      ROOT_ORDERED_TEST_PNTR IN
+                      (
+                          SELECT ROOT_ORDERED_TEST_PNTR
+                          FROM dbo.LAB_TEST ltr
+                          WHERE
+                              LAB_TEST_TYPE = 'Order'
+                            AND record_status_cd = 'INACTIVE'
+                      )
+                    AND l.record_status_cd <> 'INACTIVE'
+              );
```
ericnagel: The subquery unnecessarily re-joins LAB100 just to return RESULTED_LAB_TEST_KEY back to the outer query, which is already scanning LAB100. The outer IN (...) and the inner INNER JOIN dbo.LAB100 are doing redundant work.

```suggestion
          UPDATE l
          SET record_status_cd = 'INACTIVE'
          FROM dbo.LAB100 l
          INNER JOIN dbo.LAB_TEST lt ON lt.LAB_TEST_KEY = l.RESULTED_LAB_TEST_KEY
          WHERE l.record_status_cd <> 'INACTIVE'
          AND EXISTS (
            SELECT 1
            FROM dbo.LAB_TEST ltr
            WHERE ltr.ROOT_ORDERED_TEST_PNTR = lt.ROOT_ORDERED_TEST_PNTR
                AND ltr.LAB_TEST_TYPE = 'Order'
                AND ltr.record_status_cd = 'INACTIVE'
          );
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/371-sp_lab100_cleanup.sql
```diff
@@ -0,0 +1,192 @@
...
+          SET @PROC_STEP_NAME = 'DELETE REMOVED OBSERVATIONS FROM LAB100';
+
+          /* Remove keys in LAB100 that no longer exist in LAB_TEST. */
+          DELETE FROM dbo.lab100
+          WHERE  resulted_lab_test_key IN (
+                 SELECT DISTINCT l.resulted_lab_test_key
+                 FROM   dbo.lab100 l
+                 EXCEPT
+                 SELECT lt.lab_test_key
+                 FROM   dbo.lab_test lt);
```
ericnagel: - `DISTINCT` is redundant (`EXCEPT` already returns distinct rows)
- `LAB100` is scanned twice: once in the subquery to collect keys, and again in the outer `DELETE...WHERE IN (...)`. The `EXCEPT` also forces SQL Server to materialize both full sets before differencing them.

A `NOT EXISTS` rewrite eliminates both:

```suggestion
          DELETE l
          FROM dbo.lab100 l
          WHERE NOT EXISTS (
            SELECT 1
            FROM dbo.lab_test lt
            WHERE lt.lab_test_key = l.resulted_lab_test_key
          );
```

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/rdb.changelog-7.13.yaml
ericnagel: @ericbuckley why false? I thought we wanted to re-run these if they change? Although this particular migration should drop the index if it exists, then create it - if the file changed.

ericbuckley: I was thinking that table schema and/or data migrations should be set to false, as those should always be additive.  If we're updating a stored procedure, setting to true works for me.

mpeels: Updated `runOnChange` for the index to `false`.

PR #982, opened 2026-08-05
Err/app 932/outbreak ind empty string std investigations

eliSkylight:
## Description
Provided a fix for empty outbreak_ind_val column on several tables. The nrt_investigation table is depended on by several downstream tables to determine their outbreak_ind columns. Since the _val version of this column in nrt_investigation is an empty space instead of NULL, the downstream tables become inconsistent with MasterEtl's output for this column. The fix updates the Investigation service module to check for an empty string and set the respective field to null as needed. A SQL solution would require an inner query in sp_investigation_event which is why the current approach was taken instead. Functional tests that previously expected an empty string for this column have been updated accordingly.

## Related Issue
[APP-928](https://cdc-nbs.atlassian.net/browse/APP-928)

## Additional Notes

## Checklist


ericbuckley:
@eliSkylight regarding 

> A SQL solution would require an inner query in sp_investigation_event which is why the current approach was taken instead.

Why not do the same thing that we have for the `outbreak_ind` column?

```sql
NULLIF(results.outbreak_ind_val, '') as outbreak_ind_val,
```

eliSkylight:
> @eliSkylight regarding
> 
> > A SQL solution would require an inner query in sp_investigation_event which is why the current approach was taken instead.
> 
> Why not do the same thing that we have for the `outbreak_ind` column?
> 
> ```sql
> NULLIF(results.outbreak_ind_val, '') as outbreak_ind_val,
> ```

@ericbuckley because `outbreak_ind_val `isn't originally in the `Public_Health_Case` table, I was not confident the value would an empty string. As a result I tried doing
`NULLIF(LTRIM(RTRIM(results.outbreak_ind_val)), '') as outbreak_ind_val`. This caused an error message saying `outbreak_ind_val` is an invalid column name. A brief investigation into that showed that that extra layer causes the `results.outbreak_ind_val` to be "out-of-scope" so-to-speak. So, I went with the Java solution.

Trying the `NULLIF `causes tests to fail because it does not handle whitespace. The Java solution consistently passes tests.

reporting-pipeline-service/src/main/resources/db/changelog/migrations/v7.13/rdb/routines/056-sp_investigation_event-001.sql
ericbuckley, line 99:
```suggestion
               NULLIF(TRIM(results.outbreak_ind_val), '') as outbreak_ind_val
```

@eliSkylight is this also necessary?

eliSkylight: from my testing, no

